PR #9, opened 2021-05-17
add numpy to the installation instructions

pmeier:
The test suite depends on `numpy` since we import 

https://github.com/data-apis/array-api-tests/blob/b47e667af1f216572862bb23c8b1d5ed2e9f0a3a/array_api_tests/hypothesis_helpers.py#L9

which in turn [imports `numpy` unconditionally](https://github.com/HypothesisWorks/hypothesis/blob/a00f986c5be7890312235610ab694c97501f5a81/hypothesis-python/src/hypothesis/extra/numpy.py#L20). Thus, we should add it to the installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,11 @@ specification that are not yet tested here.
 
 To run the tests, first install the testing dependencies
 
-    pip install pytest hypothesis
+    pip install pytest hypothesis numpy
 
 or
 
-    conda install pytest hypothesis
+    conda install pytest hypothesis numpy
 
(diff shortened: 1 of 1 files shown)


asmeurer:
NumPy isn't *supposed* to be a dependency. We need to update hypothesis so that the array strategies can be array module independent. We need to do this anyway so that we can use the more powerful array generation strategies without depending on numpy. For now, though I guess we do depend on NumPy. It didn't occur to me that this would happen when I used the mutually_broadcastable_shapes strategy, but it's hard to go without that strategy.

PR #11, opened 2021-05-19
add uint vs int promotion tables to numeric input types

pmeier:


asmeurer:
Does this add anything that was not already there?

pmeier:
Yes, the mixed `uint` vs. `int` promotion tables. They are currently not covered by `"numeric"` input types, but numeric should be everything except bool, no?

asmeurer:
This is just extracting the values of those dictionaries, which should already include everything. Maybe this should be done in a less inconspicuous way.

pmeier:
:facepalm: Sorry, my bad.

PR #12, opened 2021-05-19
all -> array_module.all

pmeier:
The builtin `all` requires the object to be iterable, which 0d arrays are usually not.

```python
import numpy as np

all(np.array(True))
```

```
TypeError: iteration over a 0-d array
```


asmeurer:
I pushed an alternate fix for this.

PR #15, opened 2021-06-10
explicitly check for Python scalars in constant tests

### LINKED ISSUE #14: Test explicitly that constants are Python scalars
data-apis/array-api#169 clarified that the constants `e`, `inf`, `nan`, and `pi` should be Python scalars and not just behave like one. Our current tests

https://github.com/data-apis/array-api-tests/blob/6e3684b7948564b3ecd69f5b066aaf9c722120ec/array_api_tests/test_constants.py#L5-L11

don't account for that.

pmeier:
Fixes #14.

### DIFF
--- a/array_api_tests/test_constants.py
+++ b/array_api_tests/test_constants.py
@@ -1,9 +1,13 @@
+import math
+
 from ._array_module import (e, inf, nan, pi, equal, isnan, abs, full, float32,
                             float64, less, isinf, greater, all)
 from .array_helpers import one
 
 def test_e():
-    # Check that e acts as a scalar
+    # Check that e is a Python scalar
+    assert isinstance(e, float), "e is not a Python scalar"
+
     E = full((1,), e, dtype=float64)
 
     # We don't require any accuracy. This is just a smoke test to check that
@@ -11,7 +15,9 @@ def test_e():
     assert all(less(abs(E - 2.71), one((1,), dtype=float64))), "e is not the constant e"
 
 def test_pi():
-    # Check that pi acts as a scalar
+    # Check that pi is a Python scalar
+    assert isinstance(pi, float), "pi is not a Python scalar"
+
     PI = full((1,), pi, dtype=float64)
 
     # We don't require any accuracy. This is just a smoke test to check that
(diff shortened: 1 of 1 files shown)


rgommers:
I don't think this is a good idea, see https://github.com/data-apis/array-api/issues/154#issuecomment-833746576.

`isinstance` checks are in general very restrictive. It's hard to express how to be precise enough in text, but if it properly ducktypes at runtime, and statically types as the given type (`float` in this case), that should be considered spec-confirming. 

The same applies to shape tuples, e.g. TensorFlow has a custom `Shape` object that ducktypes as a tuple.

pmeier:
data-apis/array-api#169 clarified

> Each constant **must have a Python floating-point data type** (i.e., float) and be provided as a Python scalar value. [Emphasis mine]

So for the current spec duck typing is not allowed and the tests proposed by me should happen. We could switch the the check to [`numbers.Number`](https://docs.python.org/3/library/numbers.html#numbers.Number) to make it more permissive.

I'm fine with allowing other objects *if* there is a precise definition of what interfaces the object needs to provide to be considered compliant. IMO

> if it properly ducktypes at runtime

is too vague for that.

rgommers:
Yes, we should fix that wording.

asmeurer:
This looks good, assuming we clarify whether it should be float or Number. You can ignore GitHub actions.

BvB93:
> So for the current spec duck typing is not allowed and the tests proposed by me should happen. We could switch the the check to `numbers.Number` to make it more permissive.

Something to keep in mind is, in their current state, the `number.Number` subclasses are pretty much useless for static typing (https://github.com/python/mypy/issues/3186).

posita:
Not my fight, but I eventually settled on [this approach](https://github.com/posita/dyce/blob/master/dyce/types.py) for numerical duck-type static and runtime type-checking using `Protocol`s. (I gave up trying to use the numerical tower.)

☝️ That approach is a little convoluted, but I'm happy to shed more light, if helpful or desired. In short, most of that file is setting the stage for its [`SupportsOutcome` protocol](https://github.com/posita/dyce/blob/f679229e89c73622d80bcea074501f0b9deeb8d1/dyce/types.py#L272-L287). (An "outcome" is a library-specific term. `SupportsOutcome` attempts to describe number-like objects that can be passed to [`dyce`](https://github.com/posita/dyce/) APIs where "outcomes" are expected with a reasonable shot that things won't blow up.) In theory, `SupportsOutcome` could be used directly, even though `dyce` doesn't do that. The other things in that file basically comprise a desperate attempt to enable (semi-)performant runtime type checking around `SupportsOutcome`.

Again, I'm happy to talk shop if my technique proves useful outside my narrow corner of the Pythonverse.

posita:
I don't know if you [saw](https://github.com/python/mypy/issues/3186#issuecomment-966434911), but I pulled out a lot of my number typing work-arounds into their own package called [``numerary``](/posita/numerary). If [this journey](https://posita.github.io/numerary/latest/whytho/) lines up with your own, ``numerary`` could be an intermediary solution until python/mypy#3186 is fixed. (Alternatively, the techniques used therein might provide inspiration if you can't take a dependency.) Happy to [consult here](https://github.com/posita/numerary/discussions), if helpful.

honno:
> I don't know if you [saw](https://github.com/python/mypy/issues/3186#issuecomment-966434911), but I pulled out a lot of my number typing work-arounds into their own package called [`numerary`](/posita/numerary). If [this journey](https://posita.github.io/numerary/latest/whytho/) lines up with your own, `numerary` could be an intermediary solution until [python/mypy#3186](https://github.com/python/mypy/issues/3186) is fixed. (Alternatively, the techniques used therein might provide inspiration if you can't take a dependency.) Happy to [consult here](https://github.com/posita/numerary/discussions), if helpful.

I saw this before, awesome project! I'll at some point explore using it in a branch, and from there we can see if 1) it really can cover all the bases 2) we're happy to add the dependency.

posita:
> … I'll at some point explore using it in a branch, and from there we can see if … it really can cover all the bases … .

If you have any questions or encounter any friction or gaps, please don't hesitate to reach out. I'm eager to make sure this is useful and accessible to others, and happy to help out where I can.

honno:
Closed in favour of #52, as some of the array module usage was unnecessary after data-apis/array-api#169 (and so added unwanted dependencies).

Whilst the docs say constants are "Python floats i.e. `float`", we're just checking they instance check with `SupportsFloat`, as that allows duck typed constants which some adopters might be using. Once we get more array libraries actually adopting we can see in practice how constants work out i.e. we may be too lenient, or we may even be too strict.

Thanks @pmeier for this PR, as the introduced nan and inf test logic was useful.

(@posita so yeah for this we probably don't want `numerary` for these tests, but there's lots of instances in the test suite here where we might do well to use it. Even just performance things, and generally getting rid of numerical towers would be nice.)

PR #16, opened 2021-08-12
Updating CI runs

asmeurer:
Including adding a CI run for the NumPy implementation.

### DIFF
--- a/.github/workflows/generate_stubs.yml
+++ b/.github/workflows/generate_stubs.yml
@@ -1,0 +1,28 @@
+name: Generate Stubs
+
+on: [push, pull_request]
+
+jobs:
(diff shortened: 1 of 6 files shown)


asmeurer:
Certain NumPy tests need to be skipped as xfail. I'm not sure how I can do that. I don't want to add xfail decorators to the tests themselves. Ideally I could just list the xfail tests from the command line, but I don't see a way to do that with pytest. I asked about this on StackOverflow https://stackoverflow.com/questions/68764967/xfail-pytest-tests-from-the-command-line.

asmeurer:
The generate stubs job will start working once some upstream array-api PRs are merged.

PR #18, opened 2021-09-16
Implement `test_*_like` tests

honno:
This PR introduces use of [`hypothesis.extra.array_api`](https://hypothesis.readthedocs.io/en/latest/numpy.html#array-api) to generate (multi-dimensional) arrays for some existing tests, and enable the writing some new ones. I made a rudimentary proof-of-concept in #17, but this should be a serious attempt (and my main focus for the next two weeks).

 For now my priority is getting these tests up and running, and then iteratively I can follow the styles and idioms of the compliance suite properly.

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -3,8 +3,9 @@ from operator import mul
 from math import sqrt
 
 from hypothesis.strategies import (lists, integers, builds, sampled_from,
-                                   shared, floats, just, composite, one_of,
-                                   none, booleans)
+                                   shared, tuples as hypotheses_tuples,
+                                   floats, just, composite, one_of, none,
+                                   booleans, SearchStrategy)
 from hypothesis.extra.numpy import mutually_broadcastable_shapes
 from hypothesis.extra.array_api import make_strategies_namespace
 from hypothesis import assume
@@ -70,6 +71,8 @@ def make_dtype_pairs():
     return dtype_pairs
 
 def promotable_dtypes(dtype):
+    if isinstance(dtype, SearchStrategy):
+        return dtype.flatmap(promotable_dtypes)
     dtype_pairs = make_dtype_pairs()
     dtypes = [j for i, j in dtype_pairs if i == dtype]
     return sampled_from(dtypes)
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -6,7 +6,6 @@ from hypothesis.strategies import (lists, integers, builds, sampled_from,
                                    shared, tuples as hypotheses_tuples,
                                    floats, just, composite, one_of, none,
                                    booleans, SearchStrategy)
-from hypothesis.extra.numpy import mutually_broadcastable_shapes
 from hypothesis.extra.array_api import make_strategies_namespace
 from hypothesis import assume
 
@@ -111,29 +110,25 @@ def tuples(elements, *, min_size=0, max_size=None, unique_by=None, unique=False)
     return lists(elements, min_size=min_size, max_size=max_size,
                  unique_by=unique_by, unique=unique).map(tuple)
 
-shapes = tuples(integers(0, 10)).filter(lambda shape: prod(shape) < MAX_ARRAY_SIZE)
-
 # Use this to avoid memory errors with NumPy.
 # See https://github.com/numpy/numpy/issues/15753
-shapes = tuples(integers(0, 10)).filter(
-             lambda shape: prod([i for i in shape if i]) < MAX_ARRAY_SIZE)
+shapes = xps.array_shapes(min_dims=0, min_side=0).filter(
+    lambda shape: prod(i for i in shape if i) < MAX_ARRAY_SIZE
+)
 
-two_mutually_broadcastable_shapes = mutually_broadcastable_shapes(num_shapes=2)\
+two_mutually_broadcastable_shapes = xps.mutually_broadcastable_shapes(num_shapes=2)\
     .map(lambda S: S.input_shapes)\
-    .filter(lambda S: all(prod([i for i in shape if i]) < MAX_ARRAY_SIZE for shape in S))
+    .filter(lambda S: all(prod(i for i in shape if i) < MAX_ARRAY_SIZE for shape in S))
 
 @composite
-def two_broadcastable_shapes(draw, shapes=shapes):
+def two_broadcastable_shapes(draw):
     """
     This will produce two shapes (shape1, shape2) such that shape2 can be
     broadcast to shape1.
-
     """
     from .test_broadcasting import broadcast_shapes
-
-    shape1, shape2 = draw(two_mutually_broadcastable_shapes)
-    if broadcast_shapes(shape1, shape2) != shape1:
-        assume(False)
+    shape1, shape2 =  draw(two_mutually_broadcastable_shapes)
+    assume(broadcast_shapes(shape1, shape2) == shape1)
     return (shape1, shape2)
 
 sizes = integers(0, MAX_ARRAY_SIZE)
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -243,3 +243,9 @@ def multiaxis_indices(draw, shapes):
         extra = draw(lists(one_of(integer_indices(sizes), slices(sizes)), min_size=0, max_size=3))
         res += extra
     return tuple(res)
+
+
+shared_optional_promotable_dtypes = one_of(
+    none(),
+    promotable_dtypes(shared_dtypes),
+)
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -4,7 +4,7 @@ from math import sqrt
 
 from hypothesis.strategies import (lists, integers, sampled_from,
                                    shared, floats, just, composite, one_of,
-                                   none, booleans, SearchStrategy)
+                                   none, booleans)
 from hypothesis.extra.array_api import make_strategies_namespace
 from hypothesis import assume
 
@@ -46,9 +46,15 @@ if FILTER_UNDEFINED_DTYPES:
     boolean_dtypes = boolean_dtypes.filter(lambda x: not isinstance(x, _UndefinedStub))
     dtypes = dtypes.filter(lambda x: not isinstance(x, _UndefinedStub))
 
-shared_dtypes = shared(dtypes)
+shared_dtypes = shared(dtypes, key="dtype")
 
-def make_dtype_pairs():
+# TODO: Importing things from test_type_promotion should be replaced by
+# something that won't cause a circular import. Right now we use @st.composite
+# only because it returns a lazy-evaluated strategy - in the future this method
+# should remove the composite wrapper, just returning sampled_from(dtype_pairs)
+# instead of drawing from it.
+@composite
+def mutually_promotable_dtype_pairs(draw, dtype_objects=dtype_objects):
     from .test_type_promotion import dtype_mapping, promotion_table
     # sort for shrinking (sampled_from shrinks to the earlier elements in the
     # list). Give pairs of the same dtypes first, then smaller dtypes,
@@ -66,19 +72,12 @@ def make_dtype_pairs():
         dtype_pairs = [(i, j) for i, j in dtype_pairs
                        if not isinstance(i, _UndefinedStub)
                        and not isinstance(j, _UndefinedStub)]
-    return dtype_pairs
-
-def promotable_dtypes(dtype):
-    if isinstance(dtype, SearchStrategy):
-        return dtype.flatmap(promotable_dtypes)
-    dtype_pairs = make_dtype_pairs()
-    dtypes = [j for i, j in dtype_pairs if i == dtype]
-    return sampled_from(dtypes)
-
-def mutually_promotable_dtype_pairs(dtype_objects=dtype_objects):
-    dtype_pairs = make_dtype_pairs()
     dtype_pairs = [(i, j) for i, j in dtype_pairs if i in dtype_objects and j in dtype_objects]
-    return sampled_from(dtype_pairs)
+    return draw(sampled_from(dtype_pairs))
+
+shared_mutually_promotable_dtype_pairs = shared(
+    mutually_promotable_dtype_pairs(), key="mutually_promotable_dtype_pair"
+)
 
 # shared() allows us to draw either the function or the function name and they
 # will both correspond to the same function.
@@ -245,10 +244,10 @@ def multiaxis_indices(draw, shapes):
 
 
 shared_arrays1 = xps.arrays(
-    dtype=shared_dtypes,
+    dtype=shared_mutually_promotable_dtype_pairs.map(lambda pair: pair[0]),
     shape=shared(two_mutually_broadcastable_shapes, key="shape_pair").map(lambda pair: pair[0]),
 )
 shared_arrays2 = xps.arrays(
-    dtype=promotable_dtypes(shared_dtypes),
+    dtype=shared_mutually_promotable_dtype_pairs.map(lambda pair: pair[1]),
     shape=shared(two_mutually_broadcastable_shapes, key="shape_pair").map(lambda pair: pair[1]),
 )
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -242,9 +242,3 @@ def multiaxis_indices(draw, shapes):
         extra = draw(lists(one_of(integer_indices(sizes), slices(sizes)), min_size=0, max_size=3))
         res += extra
     return tuple(res)
-
-
-shared_optional_promotable_dtypes = one_of(
-    none(),
-    promotable_dtypes(shared_dtypes),
-)
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -2,11 +2,11 @@ from functools import reduce
 from operator import mul
 from math import sqrt
 
+from hypothesis import assume
 from hypothesis.strategies import (lists, integers, sampled_from,
                                    shared, floats, just, composite, one_of,
                                    none, booleans)
 from hypothesis.extra.array_api import make_strategies_namespace
-from hypothesis import assume
 
 from .pytest_helpers import nargs
 from .array_helpers import (dtype_ranges, integer_dtype_objects,
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -242,3 +242,13 @@ def multiaxis_indices(draw, shapes):
         extra = draw(lists(one_of(integer_indices(sizes), slices(sizes)), min_size=0, max_size=3))
         res += extra
     return tuple(res)
+
+
+shared_arrays1 = xps.arrays(
+    dtype=shared_dtypes,
+    shape=shared(two_mutually_broadcastable_shapes, key="shape_pair").map(lambda pair: pair[0]),
+)
+shared_arrays2 = xps.arrays(
+    dtype=promotable_dtypes(shared_dtypes),
+    shape=shared(two_mutually_broadcastable_shapes, key="shape_pair").map(lambda pair: pair[1]),
+)
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -130,7 +130,10 @@ sizes = integers(0, MAX_ARRAY_SIZE)
 sqrt_sizes = integers(0, SQRT_MAX_ARRAY_SIZE)
 
 # TODO: Generate general arrays here, rather than just scalars.
-numeric_arrays = builds(full, just((1,)), floats())
+numeric_arrays = xps.arrays(
+    dtype=shared(xps.floating_dtypes(), key='dtypes'),
+    shape=shared(xps.array_shapes(), key='shapes'),
+)
 
 @composite
 def scalars(draw, dtypes, finite=False):
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -2,10 +2,9 @@ from functools import reduce
 from operator import mul
 from math import sqrt
 
-from hypothesis.strategies import (lists, integers, builds, sampled_from,
-                                   shared, tuples as hypotheses_tuples,
-                                   floats, just, composite, one_of, none,
-                                   booleans, SearchStrategy)
+from hypothesis.strategies import (lists, integers, sampled_from,
+                                   shared, floats, just, composite, one_of,
+                                   none, booleans, SearchStrategy)
 from hypothesis.extra.array_api import make_strategies_namespace
 from hypothesis import assume
 
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -16,7 +16,7 @@ from .array_helpers import (dtype_ranges, integer_dtype_objects,
                             integer_or_boolean_dtype_objects, dtype_objects)
 from ._array_module import full, float32, float64, bool as bool_dtype, _UndefinedStub
 from . import _array_module
-from ._array_module import mod as xp
+from . import _array_module as xp
 
 from .function_stubs import elementwise_functions
 
@@ -49,8 +49,7 @@ if FILTER_UNDEFINED_DTYPES:
 
 shared_dtypes = shared(dtypes)
 
-@composite
-def mutually_promotable_dtypes(draw, dtype_objects=dtype_objects):
+def make_dtype_pairs():
     from .test_type_promotion import dtype_mapping, promotion_table
     # sort for shrinking (sampled_from shrinks to the earlier elements in the
     # list). Give pairs of the same dtypes first, then smaller dtypes,
@@ -60,17 +59,25 @@ def mutually_promotable_dtypes(draw, dtype_objects=dtype_objects):
     # pairs (XXX: Can we redesign the strategies so that they can prefer
     # shrinking dtypes over values?)
     sorted_table = sorted(promotion_table)
-    sorted_table = sorted(sorted_table, key=lambda ij: -1 if ij[0] == ij[1] else sorted_table.index(ij))
-    dtype_pairs = [(dtype_mapping[i], dtype_mapping[j]) for i, j in
-                   sorted_table]
-
-    filtered_dtype_pairs = [(i, j) for i, j in dtype_pairs if i in
-                            dtype_objects and j in dtype_objects]
+    sorted_table = sorted(
+        sorted_table, key=lambda ij: -1 if ij[0] == ij[1] else sorted_table.index(ij)
+    )
+    dtype_pairs = [(dtype_mapping[i], dtype_mapping[j]) for i, j in sorted_table]
     if FILTER_UNDEFINED_DTYPES:
-        filtered_dtype_pairs = [(i, j) for i, j in filtered_dtype_pairs
-                                if not isinstance(i, _UndefinedStub)
-                                and not isinstance(j, _UndefinedStub)]
-    return draw(sampled_from(filtered_dtype_pairs))
+        dtype_pairs = [(i, j) for i, j in dtype_pairs
+                       if not isinstance(i, _UndefinedStub)
+                       and not isinstance(j, _UndefinedStub)]
+    return dtype_pairs
+
+def promotable_dtypes(dtype):
+    dtype_pairs = make_dtype_pairs()
+    dtypes = [j for i, j in dtype_pairs if i == dtype]
+    return sampled_from(dtypes)
+
+def mutually_promotable_dtype_pairs(dtype_objects=dtype_objects):
+    dtype_pairs = make_dtype_pairs()
+    dtype_pairs = [(i, j) for i, j in dtype_pairs if i in dtype_objects and j in dtype_objects]
+    return sampled_from(dtype_pairs)
 
 # shared() allows us to draw either the function or the function name and they
 # will both correspond to the same function.
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -54,7 +54,7 @@ shared_dtypes = shared(dtypes, key="dtype")
 # should remove the composite wrapper, just returning sampled_from(dtype_pairs)
 # instead of drawing from it.
 @composite
-def mutually_promotable_dtype_pairs(draw, dtype_objects=dtype_objects):
+def mutually_promotable_dtypes(draw, dtype_objects=dtype_objects):
     from .test_type_promotion import dtype_mapping, promotion_table
     # sort for shrinking (sampled_from shrinks to the earlier elements in the
     # list). Give pairs of the same dtypes first, then smaller dtypes,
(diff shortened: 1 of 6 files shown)


asmeurer:
Great. Let me know if you have any questions or issues.

honno:
So currently this PR:

1. Implements `test_equal`
2. Implements `test_*_like` tests in `test_creation_functions.py`
3. Modifies `hypothesis_helpers.numeric_arrays` so it generates two nd-arrays with the same shape/dtype

Good news is that I didn't catch any bugs in `numpy.array_api` with these! (only Hypothesis heh, now patched)

You might be happy with just this to merge now? Otherwise next week I can get started going over the whole test suite to see where you'd like nd-array support.

In regards to how you've structured strategies, I didn't want to touch `hypothesis_helpers.py` too much as the conventions are a bit beyond me still. I'll give it a second look next week, or if you'd like to do it yourself then please feel free to commit to this branch. I've used `st.shared()` at some points and I imagine it'd be nicer if it's all done in `hypothesis_helpers.py` like with everything else.

Interestingly in `test_full_like` I use `shared_dtypes.flatmap(xps.from_dtype)` as the strategy for `fill_value`, as opposed to `hypothesis_helpers.scalars(shared_dtypes)`. This might be a `slices()` situation again, but I recommend you check out how I wrote `xps.from_dtype()` because it *seems* to generate all the same things, and is pretty well tested inside Hypothesis. Generally I'll need to go through the custom strategies to see what's not covered in the `hypothesis.extra.array_api` strategies.

Also, would it be appropriate to rename `numeric_arrays` to `floating_arrays`?

---

Beyond this PR please feel free to assign me any work.

I may be misremembering this, but if you wanted to use Black then obviously that could uglify some current conventions (e.g. the `assert <condition>, <msg>` pattern), so I could attempt to make that transition more palatable for you to at least give you some ideas for the future (if not merge). Generally I'd love to see a [pre-commit](https://pre-commit.com/) config—for CI you/me could also then just use `pre-commit run --all-files --show-diff-on-failure`.

I would also like a small list of "house style" things (e.g. "test function signatures should mirror the respective Array API function signatures) in the README's contributing section. To this point, I *personally* would like to see the repo use explicit namespaces for both the top-level Hypothesis strategies and the stubbed array module, i.e. `from hypothesis import strategies as st` and `from . import _array_module as xp`—I'd totally understand if you'd prefer these things be directly imported.

asmeurer:
I wouldn't say there's a house style so much as just my own style, since I'm the only author of the code here so far. I would say you're worried about style too much. I don't really care that much about the way things look beyond very basic PEP 8 (which you seem to be following just fine). For instance you use @given with keyword arguments and I generally use it with positional arguments, but either way is fine. I'm much more concerned about the behavior, like if strategies properly test the full sample space. The only somewhat stylistic thing I would say is that really long or duplicated strategies should be factored out into variables (in the hypothesis_helpers.py file) so that the @given decorators are easier to read. 

Other than that, pyflakes should pass (the linting test on CI). pyflakes only checks for logical errors, not stylistic ones. The generate stubs build should start working if you merge master. It looks like NumPy is going to fail until I update upstream with some new functions, so maybe I should add some XFAILs for now so we can make sure everything else passes. 

I'm not really a huge fan of black, so I'd prefer we don't use it. 

> I would also like a small list of "house style" things (e.g. "test function signatures should mirror the respective Array API function signatures") in the README's contributing section. To this point, I personally would like to see the repo use explicit namespaces for both the top-level Hypothesis strategies and the stubbed array module, i.e. from hypothesis import strategies as st and from . import _array_module as xp—but I'd totally understand if you'd prefer these things be directly imported.

I'm not exactly sure what you mean about the signatures thing. I agree about namespaces. I'm not sure why I did it separately for the creation function tests, but you can see the elementwise function tests do it that way. Actually there is a little bit of a distinction which maybe isn't clear, which is that some array functions are exported in array_helpers.py. These are supposed to be a subset of functions that need to be implemented for some tests to work, because they are themselves used to test other functions (the list of exported functions has unfortunately grown quite large, and we should probably try to cull it down). 

After this PR, I think the next place to update stuff is in the elementwise tests. All those tests should use full arrays. Some of the tests aren't properly written with full arrays in mind, so they will need to be updated. You might want to start from the bottom of the file (ignore the tests that aren't written yet), as I was a little more considerate of this more recently. I don't expect any major issues since any tests that already work on scalars should work on full arrays.

honno:
Ah, having spent a month getting headaches trying to adhere to ["Black++"](https://github.com/Zac-HD/shed), I didn't know what to do with myself :sweat_smile: Thanks for clarifying what you do expect!

> I'm not exactly sure what you mean about the signatures thing.

Just to clarify, this goes back to our discussion on shared strategies vs `data.draw()`, e.g. `test_full_like()` takes the parameters `a, fill_value, dtype` just like the method it's testing, `xp.full_like()` (as you said before this really helps debugging).

asmeurer:
> Just to clarify, this goes back to our discussion on shared strategies vs data.draw(), e.g. test_full_like() takes the parameters a, fill_value, dtype just like the method it's testing, xp.full_like() (as you said before this really helps debugging).

I guess I see what you mean. You might have a better idea of what these "conventions" are than I do. Stuff like using draw() is something I would never really consider doing in the first place. So maybe if you think such a note in the README would be helpful you can write up something yourself and I can let you know if they all make sense.

array_api_tests/test_creation_functions.py
```diff
@@ -77,9 +78,28 @@ def test_empty(shape, dtype):
...
-# TODO: implement empty_like (requires hypothesis arrays support)
-def test_empty_like():
-    pass
+
+@given(
+    a=xps.arrays(
+        dtype=shared_dtypes,
+        shape=xps.array_shapes(),
+    ),
+    dtype=one_of(none(), shared_dtypes),
```
asmeurer, line 87: This shouldn't be shared dtypes. The point of the dtype parameter is to specify a different dtype than the input. Actually there's still an open question right now as to whether this should be any set of dtypes or just promotable ones (https://github.com/data-apis/array-api/issues/264). 

For now, we should make `dtypes` be either None or a promotable dtype from `a.dtype`. This will require a new `promotable_dtypes` helper strategy, which would be similar to `mutually_promotable_dtypes` but instead takes a dtype as input and draws a dtype that can be promoted from that dtype in the dtype table (including itself).

honno: So there's now a `promotable_dtypes` strategy, being used in the `test_*_like` methods. Note I renamed `mutually_promotable_dtypes` to `mutually_promotable_dtype_pairs` if that's ok (to distinguish what the two strategies generate), and there's a little refactor of the promotion table things in the method `make_dtype_pairs`. Hope it's okay!

honno:
I'm happy for another review. Will likely open a (draft) elementwise PR tomorrow.

array_api_tests/meta_tests/test_hypothesis_helpers.py
```diff
@@ -0,0 +1,37 @@
...
+from .._array_module import _UndefinedStub
+from ..array_helpers import dtype_objects
+from ..hypothesis_helpers import (mutually_promotable_dtype_pairs,
+                                  promotable_dtypes)
+
+UNDEFINED_DTYPES = any(isinstance(d, _UndefinedStub) for d in dtype_objects)
+pytestmark = [pytest.mark.skipif(UNDEFINED_DTYPES, reason="undefined dtypes")]
+
+
+def test_promotable_dtypes():
```
asmeurer, line 14: I've never thought about how to test strategies before. It's probably a good idea for some of the more complex strategies. Is this how strategies are typically tested in hypothesis itself?

array_api_tests/hypothesis_helpers.py
```diff
@@ -129,8 +141,10 @@ def two_broadcastable_shapes(draw, shapes=shapes):
 
 nonbroadcastable_ones_array_two_args = hypotheses_tuples(ones_arrays, ones_arrays)
 
-# TODO: Generate general arrays here, rather than just scalars.
-numeric_arrays = builds(full, just((1,)), floats())
+numeric_arrays = xps.arrays(
```
asmeurer, line 144: Just a note here: we're going to want to split this across the different [dtype categories](https://github.com/data-apis/array-api-tests/blob/dfe977b9efc0b250bb152f9e23ed591593c3ebf2/array_api_tests/test_type_promotion.py). I guess this strategy is currently used in the special cases tests which are generated automatically, so we would have to also update the generate_stubs script to parse said categories from the spec. I think at present every function that has special cases accepts floating-point inputs which is why this works. This might end up being complicated, so it's something I'll take a look at myself at some point. We are also going to need to consider how things will work with complex dtypes once those start being added later this year.

array_api_tests/test_creation_functions.py
```diff
@@ -130,9 +150,31 @@ def test_full(shape, fill_value, dtype):
...
 
-# TODO: implement full_like (requires hypothesis arrays support)
-def test_full_like():
-    pass
+@given(
+    a=xps.arrays(
+        dtype=shared_dtypes,
+        shape=xps.array_shapes(),
+    ),
+    fill_value=shared_dtypes.flatmap(promotable_dtypes).flatmap(xps.from_dtype),
```
asmeurer, line 158: Ah, I didn't see that `from_dtype` was added. That is useful, and can probably replace some of the strategies in the hypothesis helpers (assuming it is implemented the same way).

array_api_tests/test_creation_functions.py
```diff
@@ -77,9 +78,28 @@ def test_empty(shape, dtype):
...
-# TODO: implement empty_like (requires hypothesis arrays support)
-def test_empty_like():
-    pass
+
+@given(
+    a=xps.arrays(
+        dtype=shared_dtypes,
+        shape=xps.array_shapes(),
+    ),
+    dtype=one_of(none(), shared_dtypes.flatmap(promotable_dtypes)),
```
asmeurer, line 87: It's possible I'm wrong here (flatmap always confuses me), but can't this (and elsewhere) just be `promotable_dtypes(shared_dtypes)`?

array_api_tests/test_creation_functions.py
```diff
@@ -199,9 +241,34 @@ def test_ones(shape, dtype):
...
     assert all(equal(a, full((), ONE, **kwargs))), "ones() array did not equal 1"
 
-# TODO: implement ones_like (requires hypothesis arrays support)
-def test_ones_like():
-    pass
+
+@given(
+    a=xps.arrays(
+        dtype=shared_dtypes,
+        shape=xps.array_shapes(),
```
asmeurer, line 248: I think we need to keep the shapes strategy as the one defined in hypothesis helpers. This makes it easy to globally limit the array shape, e.g., if the default strategy generates shapes that use too much memory or are too slow. It also is necessary to work around this NumPy bug https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/hypothesis_helpers.py#L102-L105. Actually, it would be much more convenient if the hypothesis array_shapes strategy had a max_size argument, which controlled the maximum prod(shape).

array_api_tests/meta_tests/test_hypothesis_helpers.py
honno: Yep, although I'll clean it up now as it could just be top-level `@given()` and `assert in ...` in the method body, instead of whatever set stuff I'm doing.

array_api_tests/test_creation_functions.py
honno: Ah I wrote to `promotable_dtypes` to only except a single dtype, not a dtype-generating strategy. Actually I'll make it accept both in it's method body.

array_api_tests/test_creation_functions.py
honno: I resolved this, thanks.

Just noting I personally wouldn't want to implement a `max_size` arg in `xps.array_shapes()`, especially if one can just use `.filter(prod(s) > max_size)`. First-party Hypothesis strategies try to avoid scenarios where a user can define "unsatisfiable" strategies like `xps.array_shapes(min_dims=5, min_side=5, max_size=24)`.

honno:
I added support for both promotable dtypes and broadcastable shapes in `test_equal`, if you'd like to take a look. I can refactor it later (would like to get a sense for how the other elementwise tests need to work before doing so).

Happy for a review again.

asmeurer:
Looks like CI is going to be failing until I can get NumPy updated. Maybe I should push a temporary fix to xfail the known tests.

array_api_tests/test_elementwise_functions.py
```diff
@@ -374,9 +374,19 @@ def test_divide(args):
...
+
+
+@given(
+    x1=xps.arrays(
+        dtype=shared_dtypes,
+        shape=shared(two_broadcastable_shapes(), key="shape_pair").map(lambda pair: pair[0])
+    ),
+    x2=xps.arrays(
+        dtype=promotable_dtypes(shared_dtypes),
+        shape=shared(two_broadcastable_shapes(), key="shape_pair").map(lambda pair: pair[1])
```
asmeurer, line 386: This is complicated enough that we should factor it out into a helper function. Maybe something like

```py
@composite
def two_arrays(draw, dtypes=dtypes):
    # return two arrays with promotable dtypes drawn from dtypes and mutually broadcastable shapes
```

and we can then use 

```py
@given(*two_arrays(dtypes)) # replacing dtypes with the appropriate strategy
```

Also note that this should be two_mutually_broadcastable_shapes, not two_broadcastable_shapes. The latter only creates shapes that broadcast to the first shape. We should probably rename it to something clearer.

array_api_tests/hypothesis_helpers.py
```diff
@@ -55,17 +58,27 @@ def mutually_promotable_dtypes(draw, dtype_objects=dtype_objects):
...
-        filtered_dtype_pairs = [(i, j) for i, j in filtered_dtype_pairs
-                                if not isinstance(i, _UndefinedStub)
-                                and not isinstance(j, _UndefinedStub)]
-    return draw(sampled_from(filtered_dtype_pairs))
+        dtype_pairs = [(i, j) for i, j in dtype_pairs
+                       if not isinstance(i, _UndefinedStub)
+                       and not isinstance(j, _UndefinedStub)]
+    return dtype_pairs
+
+def promotable_dtypes(dtype):
```
asmeurer, line 71: Drawing a dtype then using promotable_dtypes will have a different sampling distribution than mutually_promotable_dtype_pairs. I don't know if that matters. The asymmetry of `promotable_dtypes` does feel a little weird. I think we might want to instead try to use mutually_promotable_dtypes in the elementwise strategies so that they are clearer.

array_api_tests/test_creation_functions.py
```diff
@@ -130,9 +152,31 @@ def test_full(shape, fill_value, dtype):
...
 
-# TODO: implement full_like (requires hypothesis arrays support)
-def test_full_like():
-    pass
+@given(
+    a=xps.arrays(
+        dtype=shared_dtypes,
+        shape=shapes,
+    ),
+    fill_value=promotable_dtypes(shared_dtypes).flatmap(xps.from_dtype),
```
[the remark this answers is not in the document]
asmeurer: Actually, this is the case for the other _like functions as well. If dtype=None, the dtype is the same as `x`. Otherwise the dtype is `dtype`, and the dtype of `x` is irrelevant. So in all these tests, the dtype of `x` (or `a` as we're calling it here, but we should change it. I like your idea of making the test signatures exactly match the signatures in the spec) should be drawn from all dtypes and be completely independent of the `dtype` parameter. So I think actually `shared_optional_promotable_dtypes` is not needed. We can just use `arrays(dtype=dtypes)` and `dtype=one_of(none(), dtypes)`.

asmeurer:
I just noticed your strategies for the _like functions don't exactly match what the spec says, and are actually more complicated than they need to be. 

Let's go ahead and work on the elementwise stuff on a separate PR. That way we can just merge this straight away once the other issues are resolved.

asmeurer:
I updated the xfail tests on the NumPy CI so the CI should pass now if all the tests here are working properly.

array_api_tests/test_creation_functions.py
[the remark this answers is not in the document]
honno: Oh I see now! I think I've fixed this now.

Using the suggested `@given(...)` I got a falsifying example with `(x=Array([], dtype=float32), fill_value=False, dtype=None)`, so I now generate `fill_value` from `x.dtype` if `dtype == None`.

For reference this is how I interpreted the spec:
```python
>>> from numpy import array_api as xp
>>> x = xp.zeros(1, dtype=xp.uint8)
>>> xp.full_like(x, 24)       # specified behaviour
Array([24], dtype=xp.uint8)
>>> xp.full_like(x, 4.2)      # unspecified behaviour
Array([4], dtype=xp.uint8) 
>>> xp.full_like(x, 256)      # unspecified behaviour
Array([0], dtype=uint8)
```

Let me know if the updated `test_full_like` is good.

array_api_tests/test_elementwise_functions.py
honno: Good idea thanks. I think I'll revert/drop all the `test_elementwise_functions.py` changes in this PR, as it'll get annoying working on my other one.

array_api_tests/hypothesis_helpers.py
honno: Makes sense, fixed!

Just noting that if you could figure out a different place for `dtype_mapping` and `promotion_table`, you could remove the otherwise-unnecessary composite wrapper for `mutually_promotable_dtype_pairs` (see the comment I wrote).

array_api_tests/test_creation_functions.py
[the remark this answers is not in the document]
asmeurer: Here is the spec for full_like https://data-apis.org/array-api/latest/API_specification/creation_functions.html#full-like-x-fill-value-dtype-none-device-none. 

For the `dtype` parameter, it says "output array data type. If dtype is None , the output array data type must be inferred from fill_value (see full(shape, fill_value, *, dtype=None, device=None) ). Default: None ." The [`full` spec](https://data-apis.org/array-api/latest/API_specification/creation_functions.html#function-full) says "output array data type. If dtype is None , the output array data type must be inferred from fill_value . If the fill value is an int , the output array data type must be the default integer data type. If the fill value is a float , the output array data type must be the default floating-point data type. If the fill value is a bool , the output array must have boolean data type. Default: None ."

In both cases, the dtype of `x` is never mentioned. Only the value of `dtype`, or an inferred dtype from `fill_value` if `dtype=None`. So I think in your examples, the results should be 

```
>>> full_like(x, 24) # dtype=None
array(24, dtype=int64) OR array(24, dtype=int32)
>>> full_like(x, 4.2) # dtype=None
array(4.2, dtype=float64) OR array(4.2, dtype=float32)
>>> full_like(x, 2**100) # dtype=None
UNSPECIFIED
```

The dtype of `x` doesn't matter here. It could be anything. 

See https://data-apis.org/array-api/latest/API_specification/data_types.html#default-data-types. If dtype=None, the data type is the default floating-point dtype for float fill_value and the default integer dtype for int fill_value, which is either float32 or float64 and either int32 or int64, depending on the library. The only unspecified case is if the fill_value doesn't fit in the dtype. There's not really a clean way to infer the default dtype in the strategy, so we should just assume that the default is 32-bits for the purposes of generating fill_values within range when dtype=None.

So you're right that my @given isn't completely correct, because it doesn't account for unspecified fill values in the dtype=None case.

array_api_tests/hypothesis_helpers.py
asmeurer: The various data bits in the promotion test file should indeed be refactored and cleaned up. It's already used in quite a few places outside of that file. It's something I intend to do eventually, but if you see something that can help then go ahead. 

I think I also may have made a mistake by using the string values of the dtypes as the main keys in the dictionaries rather than the dtype objects themselves, leading to a lot of code where you have multiple dict lookups to get at what you want. My only concern is that the dtype objects themselves might not actually be hashable, but I think this isn't a problem in practice, so we perhaps shouldn't worry about it.

asmeurer: I opened https://github.com/data-apis/array-api-tests/issues/21 for this.

asmeurer:
I'm happy to see `promotable_dtypes` go. That name actually confused me the first time I saw it. I thought it meant dtypes promotable *from* `dtype` (upcasts), rather than *with* `dtype`.

array_api_tests/test_creation_functions.py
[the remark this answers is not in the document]
asmeurer: If you're still unsure about this, let's discuss it on Slack. I want to make sure the spec is clear about what behavior it requires and that we are matching that here.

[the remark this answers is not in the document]
honno: Thanks for your explanation! I think I understand now, and have updated the test case accordingly.

So if I'm correct in saying, this is currently a bug in NumPy:
```python
>>> from numpy import array_api as xp
>>> x = xp.zeros(5, dtype=xp.int8)
>>> x
Array([0, 0, 0, 0, 0], dtype=int8)
>>> x_like = xp.full_like(x, fill_value=False)
>>> x_like
Array([0, 0, 0, 0, 0], dtype=int8)
```
i.e. `x_like` should be `Array([False, False, False, False, False, dtype=bool)`

honno:
> I'm happy to see `promotable_dtypes` go. That name actually confused me the first time I saw it. I thought it meant dtypes promotable _from_ `dtype` (upcasts), rather than _with_ `dtype`.

Mhm I realise it's not necessary now, and definitely agree the name was confusing heh.

So all I've done today is (hopefully) fix `test_full_like()`, which I believe should be xfailed (see [comment](https://github.com/data-apis/array-api-tests/pull/18#discussion_r715047213)). Generally I'm much more comfortable with the suite now :) Just noting I'm out tomorrow, so I'll resume work on this PR (if its not in a merge-able state) and an elementwise PR next week—

array_api_tests/test_creation_functions.py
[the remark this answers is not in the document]
asmeurer: Ah you're right. This is why we have the test suite. I asked on slack to double check but it looks like this needs to be changed in NumPy. I'll make sure to include it in my next PR there.

array_api_tests/test_creation_functions.py
```diff
@@ -130,9 +157,41 @@ def test_full(shape, fill_value, dtype):
...
+
+    if dtype is None:
+        if isinstance(fill_value, bool):
+            assert x_like.dtype == xp.bool, f"{fill_value=}, but full_like() did not produce a boolean array - instead was {x_like.dtype}"
+        elif isinstance(fill_value, int):
+            assert x_like.dtype in (xp.int32, xp.int64), f"{fill_value=}, but full_like() did not produce a int32 or int64 array - instead was {x_like.dtype}"
+        elif isinstance(fill_value, float):
+            assert x_like.dtype in (xp.float32, xp.float64), f"{fill_value=}, but full_like() did not produce a float32 or float64 array - instead was {x_like.dtype}"
+        else:
+            raise Exception(f"Sanity check failed, indiciating a bug in the test suite. {fill_value=} - should be a bool, int or float")
```
asmeurer, line 185: I've been using RuntimeError for this.

asmeurer:
I noticed that we aren't tested default values for keyword arguments. I wrote up some ideas on how to do it here https://github.com/data-apis/array-api-tests/issues/22. Let me know if you have any thoughts. It isn't something we need to do here, BTW. It will be easy to add in later once we figure out a good way to do it.

honno:
So I've updated the `*_like` tests in accordance (or at least I think) with the upcoming `full_like` PR. I'm also now testing `x_like == x.dtype` when no dtype kwarg was passed.

honno:
Introduced a `kwargs` strategy as described in #22. Also updated `test_full_like` to be correct.

array_api_tests/test_creation_functions.py
```diff
@@ -94,7 +111,7 @@ def test_eye(n_rows, n_cols, k, dtype):
     else:
         a = eye(n_rows, n_cols, **kwargs)
     if dtype is None:
-        assert is_float_dtype(a.dtype), "eye() should produce an array with the default floating point dtype"
+        assert is_float_dtype(a.dtype), "eye() should returned an array with the default floating point dtype"
```
asmeurer, line 114:
```suggestion
        assert is_float_dtype(a.dtype), "eye() should return an array with the default floating point dtype"
```

array_api_tests/test_creation_functions.py
```diff
@@ -119,7 +136,7 @@ def test_full(shape, fill_value, dtype):
 
     if dtype is None:
         # TODO: Should it actually match the fill_value?
-        # assert a.dtype in _floating_dtypes, "eye() should produce an array with the default floating point dtype"
+        # assert a.dtype in _floating_dtypes, "eye() should returned an array with the default floating point dtype"
```
asmeurer, line 139:
```suggestion
        # assert a.dtype in _floating_dtypes, "eye() should return an array with the default floating point dtype"
```

array_api_tests/test_creation_functions.py
```diff
@@ -130,9 +147,32 @@ def test_full(shape, fill_value, dtype):
...
+)
+def test_full_like(x, fill_value, kw):
+    out = full_like(x, fill_value, **kw)
+    dtype = kw.get("dtype", None) or x.dtype
+    if kw.get("dtype", None) is None:
+        assert out.dtype == x.dtype, f"{x.dtype=!s}, but full_like() returned an array with dtype {out.dtype}"
+    else:
+        assert out.dtype == dtype, f"{dtype=!s}, but full_like() returned an array with dtype {out.dtype}"
+    assert out.shape == x.shape, "{x.shape=}, but full_like() returned an array with shape {out.shape}"
+    if is_float_dtype(dtype) and isnan(asarray(fill_value)):
```
asmeurer, line 171: fill_value will be a Python float so we can just check `math.isnan`. It's better to not make tests depend on other API functions when it's not necessary.

array_api_tests/test_creation_functions.py
```diff
@@ -177,54 +217,75 @@ def test_linspace(start, stop, num, dtype, endpoint):
...
+    if kw.get("dtype", None) is None:
+        assert is_float_dtype(out.dtype), "ones() returned an array with dtype {x.dtype}, but should be the default float dtype"
     else:
-        assert a.dtype == dtype
-
-    assert a.shape == shape, "ones() produced an array with incorrect shape"
-    assert all(equal(a, full((), ONE, **kwargs))), "ones() array did not equal 1"
+        assert out.dtype == dtype, f"{dtype=!s}, but ones() returned an array with dtype {out.dtype}"
+    assert out.shape == shape, "ones() produced an array with incorrect shape"
+    assert all(equal(out, full((), make_one(dtype), dtype=dtype))), "ones() array did not equal 1"
```
asmeurer, line 239: You used asarray() above but full() everywhere else. We're actually not very consistent about this in other places in the test suite, but it might be a good idea to pick only one. We want to keep the dependency graph of the tests simpler when possible, so that people can get meaningful test results if they've only implemented a subset of the API. It's not always possible, but it's something to strive for (and even the present tests don't do this as well as it could be).

array_api_tests/test_creation_functions.py
asmeurer: As long as you are cleaning up the other tests in this file you might as well implement this.

asmeurer:
This is looking good now. I only have a few minor comments. 

The CI is unfortunately bust again because I've started implemented linalg in NumPy. It might not work again until https://github.com/numpy/numpy/pull/19980 is merged, so we'll just have to check things manually. I hadn't planned on bothering updating the XFAIL list because it will all be in flux until I finish up the linalg work, but if it's an issue for you we can.

array_api_tests/test_creation_functions.py
honno: Oops I forgot about this, will do.

honno:
Great, hopefully tomorrow this PR will be up to standard.

> The CI is unfortunately bust again because I've started implemented linalg in NumPy.

Yep no problem, I'm always running pytest locally (its just pyflakes I forgot heh).

array_api_tests/test_creation_functions.py
honno: Thanks for explaining, makes sense. `asarray()` seems like easier to use, so I think I prefer that as the defacto standard for these kind of assertions.

honno:
Fixed the remaining issues I think. I didn't get round to updating the other creation tests with the `kwargs()` style and appropriate dtype assertions, not sure if that's blocking.

Was trying to get an "assertions refactor" for `test_creation_functions.py` done today to show you and see if you had any ideas—alas, next week! Kinda hard to explain without showing you, sorry. Essentially the cognitive load of writing/maintaining these few tests I've written is already making me miss things, so I wanted to try fixing that (was inspired by how Hypothesis refactors common checks).

You might of encountered this already, but just noting that `st.floats()` shrinks unexpectedly when you set bounds for it like `st.floats(-MAX_ARRAY_SIZE, MAX_ARRAY_SIZE)` (this is a known bug). There could be a quick hack for that, like `one_of(st.just(1.0), st.floats(...))`). I saw too in `mutually_promotable_dtypes` you're interested in nice minimisation behaviour. Down the line I can explore these things properly (here's the [internal util](https://github.com/HypothesisWorks/hypothesis/blob/5c48ce7a6b037a543778aa57f160a49b481d27f7/hypothesis-python/tests/common/debug.py#L36) Hypothesis uses, and an [example](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/array_api/test_from_dtype.py#L101) of its use).

asmeurer:
> Fixed the remaining issues I think. I didn't get round to updating the other creation tests with the kwargs() style and appropriate dtype assertions, not sure if that's blocking.

I actually was originally thinking the kwargs stuff should be a separate PR, so that's fine. I think this PR's already been open for too long, so let's just get it merged. 

> You might of encountered this already, but just noting that st.floats() shrinks unexpectedly when you set bounds for it like st.floats(-MAX_ARRAY_SIZE, MAX_ARRAY_SIZE) (this is a known bug).

I haven't noticed that. I know that float shrinking is weird in general because the subset of floats that can break a function is often some fractal subset. I'm not too worried about it because we have to be very minimal about tests that rely on floating point behavior, since we shouldn't assume too much about implementation details about how libraries implement certain functions.

> I saw too in mutually_promotable_dtypes you're interested in nice minimisation behaviour. Way down the line I can explore these things properly (whilst we can't use it, there's an internal util that demonstrates its possible, and an example of its use).

How does this work? The issue with the types is that the tests pick a dtype then a value for the dtype, but hypothesis doesn't know that the value picked may also have worked for a more minimal dtype. So you end up with results where the shrunk (dtype, value) is something like (uint32, 0), even though it could have been (int8, 0). I *think* this could happen even for a single dtype, although I haven't seen it yet. It's definitely worse when you draw two dtypes, because the strategies have multiple levels of dependencies on each other. If you know how to fix this that would be great.

asmeurer:
Let's merge this. If there were any other things you wanted to do here, open new PRs. 

I don't want you to have PRs open for too long, because they will get stuck in review hell and even though it's just me who's also working on this repo, it still opens the possibilities of unnecessary merge conflicts if I start touching files you're working on (if it makes your life easier, I can switch to doing my own development in PRs). So let's focus on merging stuff in even if it's only partial, as long as it works. 

I also want to make sure you aren't even blocked on my reviews, as I may not always be able to look at things in a timely fashion.

honno:
> > I saw too in mutually_promotable_dtypes you're interested in nice minimisation behaviour. Way down the line I can explore these things properly (whilst we can't use it, there's an internal util that demonstrates its possible, and an example of its use).
> 
> How does this work? The issue with the types is that the tests pick a dtype then a value for the dtype, but hypothesis doesn't know that the value picked may also have worked for a more minimal dtype. So you end up with results where the shrunk (dtype, value) is something like (uint32, 0), even though it could have been (int8, 0). I _think_ this could happen even for a single dtype, although I haven't seen it yet. It's definitely worse when you draw two dtypes, because the strategies have multiple levels of dependencies on each other. If you know how to fix this that would be great.

@asmeurer Oops forgot to respond to this. So as you might of gathered, for list-y strategies Hypothesis just see what's the first element (in this case, tuple containing dtype pair) that fails. Once we can get a test for minimisation behaviour in, it'll be easier to experiment with this (low priority for now I imagine).

PR #19, opened 2021-09-20
Use first-party Hypothesis strategies from the `st` namespace

honno:
As discussed in #18, this PR simply removes all instances of `from hypothesis.strategies import ...` in favour of `from hypothesis import strategies as st`, so that first-party strategies are used like `st.<strategy>`.

If you like, I could go over the suite to aias `_array_module` as `xp` and only call array module objects/methods with `xp.<attribute>`. Even if `_array_module` is really just a wrapper, it may add more clarity overall. Likewise I can try namespacing `hypothesis_helpers` and `array_helpers` if you'd like.

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -2,10 +2,7 @@ from functools import reduce
 from operator import mul
 from math import sqrt
 
-from hypothesis.strategies import (lists, integers, builds, sampled_from,
-                                   shared, tuples as hypotheses_tuples,
-                                   floats, just, composite, one_of, none,
-                                   booleans)
+from hypothesis import strategies as st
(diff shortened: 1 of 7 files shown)


asmeurer:
Sorry I didn't notice this PR for some reason. It has a few merge conflicts. 

Is the idea with this to make it easier to distinguish between built-in hypothesis strategies and custom ones from hypothesis_helpers?

honno:
> Is the idea with this to make it easier to distinguish between built-in hypothesis strategies and custom ones from hypothesis_helpers?

Yep. Meant it as a proof-of-concept to see how you feel. It's not a biggie!

asmeurer:
I'm really ambivalent either way. If this is something you feel would have helped you coming into the codebase then I am OK with it.

PR #23, opened 2021-09-29
Implement stubbed elementwise tests

honno:
This PR will implement the stub elementwise tests. Uses stuff from #18 so I just copied it's commit history for now—can rebase this PR to whatever merge/squash/rebase you do with that PR once it's ready.

I'll try starting from the bottom as you suggested. Right now there's just `test_trunc`, if you'd like to take a look.


array_api_tests/test_elementwise_functions.py
```diff
@@ -901,7 +902,16 @@ def test_tanh(x):
...
+@given(xps.arrays(dtype=numeric_dtypes, shape=xps.array_shapes()))
 def test_trunc(x):
-    # a = _array_module.trunc(x)
-    pass
+    a = _array_module.trunc(x)
+    assert a.dtype == x.dtype, f"{x.dtype=!s}, but trunc() did not produce a {x.dtype} array - instead was {a.dtype}"
+    if x.dtype in integer_dtype_objects:
+        assert array_all(equal(x, a)), f"{x=!s} but trunc(x)={x} - {x.dtype=!s} so trunc(x) should do nothing"
+    else:
+        # TODO: a method that generates all indices, so we don't have to flatten first
```
asmeurer, line 912: This is ndindex. See for instance test_equal.

asmeurer:
I actually meant work your way from the bottom up, fixing the existing tests (so starting with `round`). Writing new tests is a harder prospect, and will probably be easier once you see how the existing tests are written (although if you have any suggestions for improvements to the existing tests I'm all ears). 

For trunc, we the test should be similar to the tests for floor, ceil, and round (in effect, those four functions implement the four types of rounding modes, round to infinity, negative infinity, even, and zero). Actually, I think the spec text is wrong for `trunc`. It says "Rounds each element `x_i` of the input array `x` to the integer-valued number that is closest to but no greater than `x_i`." but that actually describes floor (e.g., `floor(-0.5)` is `-1.0` but `trunc(-0.5)` is `0.0`). It should say "closest to 0".

asmeurer:
Basically, I just want you to update the existing elementwise tests to accept arbitrary arrays. The ones written later (nearer the bottom of the file) will be easier to update because I wrote them with arbitrary arrays in mind, so I actually did things like used masks and iterated all indices where necessary. But for some of the tests nearer the top I did not, so they will need to be fixed to do that.

asmeurer:
> Notably I've aliased _array_module as xp, which I find significantly easier to use myself. Obviously _array_module isn't technically the array module (i.e. its stubbed and has a few goodies), so I do appreciate if you think this could be confusing. Subsequently I removed the array_helper imports which was just pointing to xp anyway. I think this is useful to distinguish what is a helper method and what is an Array API method... hopefully this is okay, not sure if I'm missing something.

The idea at least in my head was that the array_helper would be the array functions that are allowed to be used in the tests, which would then be kept as minimal as possible. It's not ended up being as minimal as I had initially hoped, so I'm not sure if it was a great idea. Although we could definitely trim it down.

honno:
> > Notably I've aliased _array_module as xp, which I find significantly easier to use myself. Obviously _array_module isn't technically the array module (i.e. its stubbed and has a few goodies), so I do appreciate if you think this could be confusing. Subsequently I removed the array_helper imports which was just pointing to xp anyway. I think this is useful to distinguish what is a helper method and what is an Array API method... hopefully this is okay, not sure if I'm missing something.
> 
> The idea at least in my head was that the array_helper would be the array functions that are allowed to be used in the tests, which would then be kept as minimal as possible. It's not ended up being as minimal as I had initially hoped, so I'm not sure if it was a great idea. Although we could definitely trim it down.

Ah I see, it makes sense now thanks. Will revert this change.

> The bitwise tests will need to just iterate over all the elements using the `ndindex` helper, similar to test_equal.

Sweet, I'll take a stab.

array_api_tests/test_elementwise_functions.py
```diff
@@ -112,20 +107,20 @@ def test_acosh(x):
     # to nan, which is already tested in the special cases.
     assert_exactly_equal(domain, codomain)
 
-@given(two_numeric_dtypes.flatmap(lambda i: two_array_scalars(*i)))
-def test_add(args):
-    x1, x2 = args
+@given(two_mutual_arrays(numeric_dtype_objects))
+def test_add(x1_and_x2):
+    x1, x2 = x1_and_x2
```
asmeurer, line 112: This is not ideal, but thinking about it, I think it's the best option. The alternative with shared stuff would just make the strategies really messy. It's really too bad that Python dropped the `def f((x, y))` syntax that used to be allowed in Python 2.

array_api_tests/test_elementwise_functions.py
```diff
@@ -851,57 +847,65 @@ def test_round(x):
...
-    # a = _array_module.trunc(x)
-    pass
+    out = xp.trunc(x)
+    assert out.dtype == x.dtype, f"{x.dtype=!s} but {out.dtype=!s}"
+    assert out.shape == x.shape, f"{x.shape} but {out.shape}"
+    if x.dtype in integer_dtype_objects:
+        assert xp.all(xp.equal(x, out)), f"{x=!s} but {out=!s}"
+    else:
+        finite_mask = xp.isfinite(out)
+        for idx in ndindex(out.shape):
```
asmeurer, line 909: You should be able to write this test without using ndindex, using the same kind of thing as test_floor and test_ceil.

asmeurer:
Why did GitHub Actions not run on this pull request?

array_api_tests/test_elementwise_functions.py
honno: Very useful, thanks.

array_api_tests/meta_tests/test_hypothesis_helpers.py
```diff
@@ -54,10 +54,11 @@ def test_kwargs():
     results = []
 
     @given(hh.kwargs(n=st.integers(0, 10), c=st.from_regex("[a-f]")))
+    @settings(max_examples=100)
```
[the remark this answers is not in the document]
honno: Yep `@settings` takes precedence over everything else. I do this so this test should pass even when we use `--max-examples=1`. This is actually what a few tests in Hypothesis does, to prevent false-negatives.

array_api_tests/test_elementwise_functions.py
honno: This works great, thanks!

array_api_tests/meta_tests/test_hypothesis_helpers.py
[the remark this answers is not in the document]
asmeurer: I don't understand why you would ever pass max-examples=1. And even if you did, why wouldn't want it to apply to this test? I typically use max-examples to increase the number of examples to do a more rigorous test run.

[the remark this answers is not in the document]
honno: Hmm, a few times now I've run `pytest --max-examples=1` for the whole suite, so I thought forcing `max_examples` here would reduce noise for myself and generally prevent any noisy false-negatives if someone else had the same idea. Generally I think it clarifies that this test requires multiple runs. I imagine both of these reasons are why Hypothesis does this internally too.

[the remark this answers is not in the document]
honno: How this `test_kwargs` test method tries to find that certain things will be generated is essentially what `find_any` does in their own test suite.

PR #26, opened 2021-10-05
Clean up the data in test_type_promotion.py

### LINKED ISSUE #21: Clean up the data in test_type_promotion.py
Everything that's not a test in test_type_promotion.py is a mess right now. There are tons of data dictionaries and helper functions. Quite a few of them are needed throughout the test suite, so they should be moved to a more central location. Furthermore, the data has a lot of redundancies, and even so, to use it, you often have to do tons of dict lookups to get at what you want. I think the biggest problem here is that I used the string values of the dtypes as the main key in the dictionaries, meaning you always have to translate those to actual dtype objects to use them. We should just use the actual dtype objects instead. The only concern here is that the dtype objects might not be hashable. But I don't think this is an actual problem in practice. If it somehow does become a problem for some library, we can worry about it then, and whether it should be worked around in the suite or if we should just require hashability in the spec (we already require equality). Finally, the parameterization logic for the tests that are in test_type_promotion is a bunch of hard to read list comprehensions. They may become easier to read once we refactor the data. If not, we should clean them up as well. So to summarize

- [ ] Move the reusable data and helper functions to a central location (I'm thinking a new file, like dtype_helpers.py). The input and output type mappings should also be moved, as they are also useful outside of just these tests.
- [ ] Change the data to be keyed on the dtype objects themselves. I don't actually care if we keep the "i4" type names anywhere. Those are used in the spec, but they aren't actually that clear. I would rather we just use the dtype name itself, like "int32", everywhere, even in the places where the string names are currently used (I think it's only in the type promotion parameterization keys).
- [ ] Other refactors like renaming variables to better names are also welcome.
- [ ] Clean up the parameterization logic in test_type_promotion.py

honno:
This will resolve #21 .

- [x] Move the reusable data and helper functions to a central location
- [ ] Change the data to be keyed on the dtype objects themselves.... or maybe dtype string names where appropriate
- [ ] Review more refactors and other niceties
- [ ] Clean up the parameterization logic in `test_type_promotion.py`


asmeurer:
Your approach to cleaning up the paramerizations seems pretty reasonable. Doing it all in a function with explicit loops instead of list comprehensions, and generating param objects instead of separate parameters and ids does look better.

array_api_tests/dtype_helpers.py
```diff
@@ -0,0 +1,289 @@
...
+    (xp.int64, xp.uint32): xp.int64,
+    # floats
+    (xp.float32, xp.float32): xp.float32,
+    (xp.float32, xp.float64): xp.float64,
+    (xp.float64, xp.float64): xp.float64,
+}
+promotion_table = {
+    (xp.bool, xp.bool): xp.bool,
+    **_numeric_promotions,
+    # TODO: dont unpack pairs of the same dtype
```
asmeurer, line 89: This doesn't matter. The dict will already de-duplicate. The only way this might matter is if we want to make the order reasonable, but in that case the below dict should be mixed with this one.

array_api_tests/test_type_promotion.py
```diff
@@ -1,738 +1,242 @@
...
+        )
+
+
+@pytest.mark.parametrize('expr, in_dtypes, out_dtype, x_filter', gen_op_params())
+@given(data=st.data())
+def test_operator_returns_array_with_correct_dtype(
+    expr, in_dtypes, out_dtype, x_filter, data
+):
+    if len(in_dtypes) == 1:
+        x = data.draw(
```
asmeurer, line 150: Is it possible to avoid `draw` here, or can strategies not access parameterization args?

honno: Yeah I'm pretty sure we have to draw dynamically, and if not then it'd probably be too much of a hack anyway. Also the labels should make it a bit clearer what arrays exactly cause problems when Hypothesis finds failures.

honno:
I think `dtype_helpers.py` now holds all the related dtype helpers.

I've attempted to refactor the scalar promotion tests. I'm not quite finished with it yet, but you might want to take a look.

Notably I introduce an "elements factories" concept to prevent triggering undefined overflow behaviours in the test code, I pass `<min/max>_value` to the underlying `xps.from_dtype()` strategy used by `xps.arrays()` via the `elements` kwarg, i.e. `xps.arrays(..., elements={...})`. This `elements` dict depends on the generated scalars, thus I use factory methods to create the appropriate  `<min/max>_value` kwargs on runtime. It's messy, but it does seem to work nicely when I look at `pytest -s --hypothesis-verbosity=verbose`... will need to sleep on it, interested to hear your thoughts. Will want to write appropriate comments too.

I also don't test NaNs and inf scalars, as I expect you want to keep that all in the generated special cases tests.

asmeurer:
> I also don't test NaNs and inf scalars, as I expect you want to keep that all in the generated special cases tests.

No we should keep those. The special case tests only test the functions, not the operators. And they also don't test the dtype, only the value.

asmeurer:
> I believe you initially desired float first, then int-uint-bool. Personally I prefer the opposite scenario, which is also how hypothesis.extra.numpy/hypothesis.extra.array_api does it. I think (bool, bool) better indicates errors which are not related to dtypes at all, just an array of any dtype. (uintX, uintX) likewise better indicates problems with numerical arrays generally, as floats obviously have a larger "elements space" of real numbers, NaNs and infs.

Sounds reasonable. It isn't important, but I also wonder if we should prefer the "default floating-point/integer dtype" first. I find it a little confusing that hypothesis always shrinks to float32. In my experience actually if only one of float32 or float64 fails, then float32 is the one that fails, because it is more likely to suffer bugs from overflows or upcasting. If anything, seeing "float32" makes me suspect that the shorter type is somehow relevant (although I know that this is general sort of problem with hypothesis shrinking that you just have to be on guard about, see https://hypothesis.works/articles/threshold-problem/). If the floating-point dtype is irrelevant to the failure, I think would generally prefer to see the default floating-point dtype (which is float64 for NumPy, but can be either in the spec). 

> Also, I can't actually test minimisation behaviour easily without using the MPL-licensed code Hypothesis does for this. I'll definitely mull it over though, so maybe expect a PR in the future for this.

Can we just import the internal hypothesis functions?

asmeurer:
Great, this works against plain `numpy` now. When I test it, I get parameter names like `array_api_tests/test_type_promotion.py::test_func_returns_array_with_correct_dtype[abs(<class 'numpy.int8'>) -> <class 'numpy.int8'>]`. The `<class 'numpy.int8'>` bit also appears in the error messages, but I think that's fine. But I think we should make the parameterization always give `abs(int8) -> int8` like it does with numpy.array_api.

I was worried what the error messages would look like with so many @composite strategies, but it looks like they are usable. Here is an example error from `numpy` (which does incorrect type promotion on 0-d arrays)

```
func = <ufunc 'add'>, in_dtypes = (<class 'numpy.int64'>, <class 'numpy.int32'>), out_dtype = <class 'numpy.int64'>
x_filter = <function <lambda>.<locals>.<lambda> at 0x112c7c0d0>, data = data(...)

    @pytest.mark.parametrize('func, in_dtypes, out_dtype, x_filter', gen_func_params())
    @given(data=st.data())
    def test_func_returns_array_with_correct_dtype(
        func, in_dtypes, out_dtype, x_filter, data
    ):
        if len(in_dtypes) == 1:
            x = data.draw(
                xps.arrays(dtype=in_dtypes[0], shape=hh.shapes).filter(x_filter), label='x'
            )
            out = func(x)
        else:
            arrays = []
            shapes = data.draw(
                hh.mutually_broadcastable_shapes(len(in_dtypes)), label='shapes'
            )
            for i, (dtype, shape) in enumerate(zip(in_dtypes, shapes), 1):
                x = data.draw(
                    xps.arrays(dtype=dtype, shape=shape).filter(x_filter), label=f'x{i}'
                )
                arrays.append(x)
            try:
                out = func(*arrays)
            except OverflowError:
                reject()
>       assert out.dtype == out_dtype, f'{out.dtype=!s}, but should be {out_dtype}'
E       AssertionError: out.dtype=int32, but should be <class 'numpy.int64'>
E       assert dtype('int32') == <class 'numpy.int64'>
E         +dtype('int32')
E         -<class 'numpy.int64'>

array_api_tests/test_type_promotion.py:99: AssertionError
------------------------------------------------------------------------------ Hypothesis -------------------------------------------------------------------------------
Falsifying example: test_func_returns_array_with_correct_dtype(
    data=data(...),
    func=<ufunc 'add'>,
    in_dtypes=(numpy.int64, numpy.int32),
    out_dtype=numpy.int64,
    x_filter=<function array_api_tests.test_type_promotion.<lambda>.<locals>.<lambda>>,
)
Draw 1 (shapes): ((), (1,))
Draw 2 (x1): array(0)
Draw 3 (x2): array([0], dtype=int32)
```

The input arrays, including shapes and dtypes are both easy to see. My only gripe is I would include the name of the function in the assertion error message.

PR #28, opened 2021-10-14
Add `pre-commit` config for pyflakes

honno:
I keep on forgetting to run `pyflakes` before pushing, so I thought I'd use [`pre-commit`](https://pre-commit.com/). Here's a minimal config for the tool. As there doesn't seem to be a `pre-commit` hook for `pyflakes` itself, I just use `flake8` and run only the pyflake-related checks via `flake8 --select F`. I've also updated the linting workflow to use `pre-commit`, to keep consistency (although I doubt `flake8` versions will change the pyflake linting behaviour).

It is "another thing", so I understand if you're not interested in merging. For the future I was thinking it might be nice to create a custom `pre-commit` hook that ensured any staged changes to `./generate_stubs.py` would trigger a check to see the related files have been updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+-   repo: https://github.com/pycqa/flake8
+    rev: '4.0.1'
+    hooks:
+    -   id: flake8
+        args: [--select, F]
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+-   repo: https://github.com/pycqa/flake8
+    rev: '4.0.1'
```
asmeurer, line 3: I'd prefer to keep this unpinned. Whenever pyflakes is updated it may create new failures, but those are also things we'd want to know about. Also we are going to want to add 3.10 testing pretty soon, and I don't know if this versions supports it yet.

PR #29, opened 2021-10-14
Skip extension tests if the respective extension is not found or is passed via `--disable-extensions`

### LINKED ISSUE #25: Clean way to optional enable/disable extensions
Now that we are adding more tests for the linalg extension, and we will soon have other extensions as well, we need to have a clean way for modules to not implement these extensions and not fail the test suite. Some test files mix extension and non-extension tests, e.g., the signature tests currently test all signatures. 

This is closely related to #20.

honno:
Should resolve #25. My current idea is to add a pytest options `--extensions`, where the user can specify any extensions they want to test alongside the normal tests (i.e. by default, no extensions are tested). 

This works by catching a custom marker `xp_extension(<ext>)` in pytest's collection hook—if the user has has passed the respective `<ext>` then the test is run, otherwise it's skipped.  For parameters it can just be passed in the `marks` kwarg e.g. `pytest.param(..., marks=xp_extension("linalg"))`, and for module level it can be defined in a file's top-level variable `pytestmark` e.g.:
```python
pytestmark = [pytest.mark.xp_extension("linalg")]
```

I think I understand how `test_signatures.py` works now, so I tomorrow I should have a ready solution to easily opt-in the `linalg` tests there and in `test_linalg.py`.

### DIFF
--- a/.github/workflows/numpy.yml
+++ b/.github/workflows/numpy.yml
@@ -37,19 +37,10 @@ jobs:
         array_api_tests/test_signatures.py::test_function_positional_args[__dlpack__]
         array_api_tests/test_signatures.py::test_function_positional_args[__dlpack_device__]
         array_api_tests/test_signatures.py::test_function_positional_args[from_dlpack]
+        array_api_tests/test_signatures.py::test_function_positional_args[to_device]
         array_api_tests/test_signatures.py::test_function_keyword_only_args[__dlpack__]
-        # Updates to the spec since the last change to numpy.array_api
-        # These will fail until NumPy is updated
-        array_api_tests/test_signatures.py::test_has_names[__index__]
-        array_api_tests/test_signatures.py::test_has_names[to_device]
-        array_api_tests/test_signatures.py::test_has_names[mT]
-        array_api_tests/test_signatures.py::test_has_names[tril]
-        array_api_tests/test_signatures.py::test_has_names[triu]
-        array_api_tests/test_signatures.py::test_has_names[matrix_transpose]
-        array_api_tests/test_signatures.py::test_has_names[permute_dims]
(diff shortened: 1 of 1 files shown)


asmeurer:
Looks like this has conflicts now.

honno:
@asmeurer fixed, fortunately just a rebase did the trick.

asmeurer:
The GitHub actions workflow should be appending to conftest (it uses `>>`). So I think it ought to work. 

If you have any suggestions for a better way to do it, let me know. I didn't get any good answers at https://stackoverflow.com/questions/68764967/xfail-pytest-tests-from-the-command-line. Maybe we should open a feature request for pytest to make something like this easier. Ideally you could just do `pytest --xfail-tests ...`.

array_api_tests/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+from pytest import mark
+
+from . import _array_module as xp
+from ._array_module import _UndefinedStub
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        '--ext',
+        '--disable-extension',
```
asmeurer, line 12: I guess if you add options in a module's conftest.py they don't show up in `pytest -h`.

README.md
```diff
@@ -41,6 +45,29 @@ import numpy as array_module
...
+Simply run the `pytest` command to run the whole test suite.
+
+```bash
+pytest
+```
+
+The test suite tries to logically organise its tests so you can find specific
+test cases whilst developing something in particular. So to avoid running the
+rather slow complete suite, you can specify particular test cases like any other
+test suite.
```
asmeurer, line 59: I just pushed a change to the README that adds a "configuring the tests" section talking about `--max-examples`. Maybe this text would make more sense there.

PR #30, opened 2021-10-19
More promotion test cases

### LINKED ISSUE #24: Add type promotion tests for non-elementwise functions
Currently the type promotion tests only test elementwise functions and array operators. But there are several other functions in the spec that also participate in type promotion. We should add them to the tests. This should be a question of just going through the parameterizations and adding any function not in the elementwise section that should participate in type promotion. We will need to make some minor generalizations to the tests as some functions only type promote on certain arguments (for example [where](https://data-apis.org/array-api/latest/API_specification/searching_functions.html#where-condition-x1-x2) has three arguments but only type promotes on the second and third). It may be a good idea to look at #21 before or concurrently with this.

honno:
This should help resolve #24. Starts from the unmerged #26 so the commit history is currently messy. I'm going to try cover some more methods before seeing about a refactor.

### DIFF
--- a/.github/workflows/numpy.yml
+++ b/.github/workflows/numpy.yml
@@ -41,6 +41,12 @@ jobs:
         array_api_tests/test_signatures.py::test_function_keyword_only_args[__dlpack__]
         # floor_divide has an issue related to https://github.com/data-apis/array-api/issues/264
         array_api_tests/test_elementwise_functions.py::test_floor_divide
+        # mesgrid doesn't return all arrays as the promoted dtype
+        array_api_tests/test_type_promotion.py::test_meshgrid
+        # https://github.com/numpy/numpy/pull/20066#issuecomment-947056094
+        array_api_tests/test_type_promotion.py::test_where
+        # shape mismatches are not handled
+        array_api_tests/test_type_promotion.py::test_tensordot
 
         EOF
 
--- a/conftest.py
+++ b/conftest.py
@@ -102,5 +102,6 @@ def pytest_collection_modifyitems(config, items):
         except StopIteration:
             pass
         # workflow xfail_ids
-        if item.nodeid in xfail_ids:
-            item.add_marker(mark.xfail(reason='xfails.txt'))
+        for id_ in xfail_ids:
+            if item.nodeid.startswith(id_):
+                item.add_marker(mark.xfail(reason='xfails.txt'))


array_api_tests/hypothesis_helpers.py
```diff
@@ -113,15 +113,21 @@ def tuples(elements, *, min_size=0, max_size=None, unique_by=None, unique=False)
...
 # Use this to avoid memory errors with NumPy.
 # See https://github.com/numpy/numpy/issues/15753
-shapes = xps.array_shapes(min_dims=0, min_side=0).filter(
-    lambda shape: prod(i for i in shape if i) < MAX_ARRAY_SIZE
-)
+def shapes(**kw):
+    if 'min_dims' not in kw.keys():
+        kw['min_dims'] = 0
+    if 'min_side' not in kw.keys():
+        kw['min_side'] = 0
```
asmeurer, line 120: You can use `kw.setdefault('min_side', 0)`.

asmeurer:
So it looks like none of the functions outside of elementwise are actually testable with the same test as elementwise. I wonder if we should just test the type promotion for these functions in their normal function tests, like I've done here https://github.com/data-apis/array-api-tests/blob/9fbe990fa9f10535b68a988585c93c25f1ae3a26/array_api_tests/test_linalg.py#L274. For matmul especially I don't know if it's worth it trying to come up with an input strategy for it just so that we can split the type promotion out into a separate test. The primary benefit of testing type promotion separately, as I understand it, is that the types are parameterized, so that you always have every combination tested. I don't know hypothesis internals enough to know how high you have to set the max examples to ensure that for the more general test, but I think it will be good in general for library authors to do a run with a high max-examples once they feel their library is confirming just to confirm the tests didn't miss anything.

array_api_tests/pytest_helpers.py
```diff
@@ -33,3 +38,22 @@ def doesnt_raise(function, message=''):
...
 def nargs(func_name):
     return len(getfullargspec(getattr(function_stubs, func_name)).args)
+
+
+def assert_dtype(
+    func_name: str,
+    in_dtypes: Tuple[DataType, ...],
+    out_name: str,
+    out_dtype: DataType,
+    expected: DataType
```
asmeurer, line 48: This function feels like it has too many arguments. `expected` could just default to taking the result type of the in_dtypes. Is the out_name really needed either?

PR #31, opened 2021-10-21
Bound `test_full` fill values to default dtypes

honno:
Currently our test suite can fail for `test_full` (depending on the Hypothesis seed or max examples) because I didn't bound `fill_value` generation to the bounds of the default dtypes. This PR fixes that by inferring the default int and float in `dtype_helpers`, and using that to prevent passing dtypes with values that cannot be represented by the default dtypes.

This should be useful for the future too.

### DIFF
--- a/array_api_tests/dtype_helpers.py
+++ b/array_api_tests/dtype_helpers.py
@@ -1,6 +1,8 @@
+from warnings import warn
 from typing import NamedTuple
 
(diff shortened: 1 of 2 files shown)


asmeurer:
Looks good. It would be nice if we had a library that defaulted to float32 to test this against. Does NumPy default to 32-bit types when compiled 32-bit?

PR #32, opened 2021-10-22
Refactor assertions in `test_creation.py`

honno:
In #30 I'm experimenting with the `ph.assert_dtype()` assertion util. As well as those promoted dtype checks, I realise a lot of dtype assertions relate to the default int and float, so I wanted to experiment with refactoring those too... I didn't get round to that today, seeing as how tricky just understanding `test_arange` was (I think it covers more now at least).

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -1,25 +1,24 @@
-from functools import reduce
-from operator import mul
-from math import sqrt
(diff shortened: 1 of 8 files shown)


asmeurer:
I've merged this. We might still need to do some work on the arange test, but for now I think it's OK. The meshgrid change is fine. We can expand it when we convert it into the actual meshgrid test.

Issue #33, opened 2021-10-27
Make two_mutual_arrays accept a dtypes strategy as input


asmeurer:
two_mutual_arrays accepts a list of dtype objects rather than a strategy for generating dtypes. This is somewhat annoying, as everything else uses the dtypes strategies.

A issue with this is that it currently ignores the filtering that is done in the dtype strategies. This makes any test using it fail in pytorch, which doesn't have uint8. 

If we can make it accept a dtypes strategy, that would be great. We may just need to have an internal mapping of dtype strategies to promotable dtype strategies. Or maybe we can just draw a dtype then draw another dtype that is promotable compatible with it.

asmeurer:
Let's leave this open for making two_mutual_arrays accept a strategy as input. It's something that confuses me almost every time I run into it, so it would be good to change it if at all possible.

honno:
Hmm, so I can't figure out how to do this nicely. Both the dtype strategies in `hypother_helpers.py` and in `xps.*` return generic strategies with seemingly no way (definitely not public) to distinguish between them (and thus map them to their respective dtype objects).

I wrote #36 at least to raise a consistent error.

honno:
I still don't think this is feasible/worth doing, and hopefully my custom error and more experience mitigates the problem you were having.

PR #34, opened 2021-10-28
Filter undefined dtypes in `hh.mutually_promotable_dtypes()`

honno:
If `max_size` is not its default value `2`, then `hh.mutually_promotable_dtypes()` before returned undefined dtypes. This PR fixes that. Related to #33.

I now bind `xps` to the underlying array module `mod`, instead of the top-level stubbed one. Hypothesis gives nice errors when it encounters erroneous situations due to partial/incorrect Array API adoption, and like with the dtype strategies in `hypothesis_helpers.py` it will work even if not all dtypes are available.

I also add two test cases, one mocking an array module with undefined dtypes, and one actually testing an array module with undefined dtypes in `test_partial_adopters.py`. Because of how the codebase is structured, it's pretty awkward to test partial adopters like [I had done in Hypothesis](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/array_api/test_partial_adoptors.py), but a liberal use of `pytest.mark.skipif()` allows us to test how our own  strategies work with different array modules.

### DIFF
--- a/array_api_tests/__init__.py
+++ b/array_api_tests/__init__.py
@@ -1,6 +1,10 @@
 from hypothesis.extra.array_api import make_strategies_namespace
 
-from ._array_module import mod as xp
+from ._array_module import mod as _xp
 
 
-xps = make_strategies_namespace(xp)
+xps = make_strategies_namespace(_xp)
+
+
+del _xp
+del make_strategies_namespace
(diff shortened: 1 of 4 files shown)


asmeurer:
I've asked Ralf to give you push access to this repo. You should probably still do PRs for now, but this will let you merge them once they are ready.

This PR looks fine to me. Feel free to merge it once you resolve the conflicts.

PR #35, opened 2021-10-29
Operator tests

### LINKED ISSUE #27: Tests for operators
Every `__operator__` method has a corresponding function which should have the same behavior, meaning the tests for those operators should be able to just reuse the test for the function. We should figure out a clean way to do this. Maybe it's just a question of parameterizing over the function so that something like `test_add` will parameterize over `add()`, `x1.__add__` and `x1.__radd__`. There's also the complication of the in-place operators (`x1.__iadd__`), which work just like the operators except the input shape and dtypes are more restricted (the result shape and dtype must be the same as the lhs). This might just be a case of only testing `op=` when the inputs would be legal. 

These are already split out in the type promotion tests (which are currently the only tests that test the operator methods). I don't know if those tests also need to be merged. If they can be merged simply they can be, but otherwise it is fine to keep them separate as they are now.

honno:
Should resolve #27, if partially. Commits start from #32.

Currently just parametrizes expressions to evaluate. As expected, testing the inplace operators require more thought.


array_api_tests/test_elementwise_functions.py
```diff
@@ -24,106 +25,106 @@
...
-    a = xp.abs(x)
-    ph.assert_shape("abs", a.shape, x.shape)
-    assert ah.all(ah.logical_not(ah.negative_mathematical_sign(a))), "abs(x) did not have positive sign"
+def test_abs(expr, x):
+    if x.dtype in dh.int_dtypes:
+        # abs of the smallest representable negative integer is not defined
+        m, _ = dh.dtype_ranges[x.dtype]
+        mask = xp.not_equal(x, ah.full(x.shape, m, dtype=x.dtype))
+        x = x[mask]
+    out = eval(expr)
```
asmeurer, line 36: It would be better to explicitly pass a globals namespace to eval, like `eval(expr, {'xp': xp})`. That way any names that are imported or defined in this file can't accidentally affect it.

asmeurer:
We could test the operator methods like `x1.__add__(x2)` instead of `x1 + x2`. That may remove the need to do `eval`. In particular, we can use the [operator](https://docs.python.org/3/library/operator.html) module to treat `x1 + x2` as the function `operator.__add__`. 

It would also make it possible to test the right operators like `__radd__`, which aren't otherwise easy to test (although to be fair, this also implies that these methods aren't actually that important).

honno:
(I've come up with a solution, so now there's full test coverage of the respective standard and inplace operators of the elementwise tests, as well as scenarios where scalars are in play... it's not worth reviewing yet though as things are still in flux)

asmeurer:
Sorry I've fallen a bit behind in reviewing. Is this PR ready for review and merge?

asmeurer:
Why was the broadcasting test moved to `meta/`? It seems to me that it's an actual test of array API module behavior, not a meta test.

honno:
> Sorry I've fallen a bit behind in reviewing. Is this PR ready for review and merge?

Yep!

> Why was the broadcasting test moved to `meta/`? It seems to me that it's an actual test of array API module behavior, not a meta test.

Ah IIRC it was actually testing your custom implementation of the broadcasting algo from the spec, so I moved it there to clarify that.

asmeurer:
I implemented the algorithm from the spec and use it as a baseline to test the array API library. I suppose we could split out the testing of the algorithm itself into meta tests, but the main point is to test the broadcasting behavior in the array library. 

To be sure, this test is less important now that we actually have tests for the actual functions. So maybe it should be refactored. A line in each multiarg function test like `assert res.shape == _broadcast_shapes(x1.shape, x2.shape)` is basically equivalent. Unlike type promotion, there may not be much benefit to having it separate once the actual function tests exist.

honno:
> To be sure, this test is less important now that we actually have tests for the actual functions. So maybe it should be refactored. A line in each multiarg function test like `assert res.shape == _broadcast_shapes(x1.shape, x2.shape)` is basically equivalent. Unlike type promotion, there may not be much benefit to having it separate once the actual function tests exist. Perhaps this was your reasoning for moving things around?

It wasn't, I just made a mistake haha. But now that you mention that, we are indeed testing broadcasting for all the related functions (or should be), so I believe we don't need it anymore. I've just removed the method. Whilst parametrized tests for broadcasting is nice, I think it's the right decision not to specifically test broadcasting (just like we removed those complicated `test_type_promotion.py` tests). You think that's okay?

asmeurer:
Yes, this sounds good. In the future, we might have to think about if we need to split tests up to help implementers who may want to test certain things even when other things aren't working. But that's something we should think about only once it becomes an issue. We may find that the most practical method for implementers will be to just directly comment out broken bits to see if other parts work. For now the test suite is heavily biased toward conforming or nearly conforming libraries, and it will be difficult to make it otherwise without some specific example to work against.

asmeurer:
Merging this so that the other PR based on it is easier to review.

PR #38, opened 2021-11-08
Test operator special cases

### LINKED ISSUE #37: Generate test cases for the special case behaviour of all operators
(This issue is a bit premature as data-apis/array-api#299 hasn't gone through—just didn't want to forget!)

honno:
Resolves #37. Creates `test_dunder_<op>.py` files for each operator with special cases to test, much like how the elementwise cases are generated.

It seems like the only problem with `numpy.array_api` (when running 2000 examples) is in-place addition.
```python
>>> x1 = xp.asarray(-0.0)
>>> x1
Array(-0., dtype=float64)
>>> x2 = xp.asarray(-0.0)
>>> xp.add(x1, x2)
Array(-0., dtype=float64)
>>> x1 + x2
Array(-0., dtype=float64)
>>> x1 += x2
>>> x1
Array(0., dtype=float64)  # not negative 0, like it is with xp.add() and __add__()
```
Interestingly I had to run like 200 examples to get this error consistently. Something for the future would be to filter strategies from the `@given()` level so that only arrays with NaNs/infs/whateverproperty-we're-testing are generated, which would mitigate this problem.


asmeurer:
Nice. Why are there additions to non-operator methods here too? Were some tests being omitted previously for some reason? 

I hope the generate_stubs script wasn't too hard to work with. That code is a bit hairy.

asmeurer:
Oh I see there were several changes in the spec recently. Let me just run generate_stubs on main so that the diff here is easier to read.

asmeurer:
I'm not sure when this started happening, but several atan and atan2 special case tests involving π are failing because they don't compare exactly equal to the given value of π for float32. The spec only requires an approximation, so this should be fixed, but I'm a little curious how I haven't seen this before.

honno:
> I hope the generate_stubs script wasn't too hard to work with. That code is a bit hairy.

Haha bit challenging at first, but it does all makes sense. Some small areas I could factor out, but IMO the only thing that was difficult to grasp was how `main()` loops through the spec files, which I could factor out in in a future PR.

> I'm not sure when this started happening, but several atan and atan2 special case tests involving π are failing because they don't compare exactly equal to the given value of π for float32. The spec only requires an approximation, so this should be fixed, but I'm a little curious how I haven't seen this before.

Couldn't reproduce on my machine, though I was out of sync of NumPy upstream or your PR branches. I can explore this but prob worthwhile in a future PR given the awkwardness of reproducing (e.g. building NumPy versions).

asmeurer:
> Couldn't reproduce on my machine, though I was out of sync of NumPy upstream or your PR branches. I can explore this but prob worthwhile in a future PR given the awkwardness of reproducing (e.g. building NumPy versions).

I think it must be a hypothesis random failure that my machine started finding. Probably will catch them if you run `pytest -k atan --max-examples=10000` (and yes, definitely not for this PR, if that wasn't clear).

generate_stubs.py
```diff
@@ -29,6 +31,50 @@
...
+    '__ne__',
+    '__neg__',
+    '__or__',
+    '__pos__',
+    '__pow__',
+    '__rshift__',
+    '__sub__',
+    '__truediv__',
+    '__xor__'
+]
```
asmeurer, line 61: Maybe not super important since these are unlikely to change, but can't we extract this list from the spec?

asmeurer:
This looks good to me. You can merge when you are happy with it.

generate_stubs.py
honno: Seemingly not, as there needs to be a way to distinguish standard operators to in-place operators.

PR #39, opened 2021-11-09
Manipulation and statistical tests, smoke all untested functions

honno:
This PR should result in at least minimal coverage for every function in the spec. My idea is that smoking the functions is easy (ignoring complicated kwarg for now), and personally this process lets me pick out what tests I'm comfortable in developing for the near-future. Commits start from #35.

### DIFF
--- a/array_api_tests/meta/test_utils.py
+++ b/array_api_tests/meta/test_utils.py
@@ -1,6 +1,8 @@
 import pytest
 
+from .. import array_helpers as ah
 from ..test_creation_functions import frange
+from ..test_manipulation_functions import axis_ndindex
 from ..test_signatures import extension_module
 from ..test_statistical_functions import axes_ndindex
 
@@ -25,6 +27,30 @@ def test_extension_func_is_not_extension():
 def test_frange(r, size, elements):
     assert len(r) == size
     assert list(r) == elements
+
+
+@pytest.mark.parametrize(
+    "shape, expected",
+    [((), [()])],
+)
+def test_ndindex(shape, expected):
+    assert list(ah.ndindex(shape)) == expected
+
+
+@pytest.mark.parametrize(
+    "shape, axis, expected",
+    [
+        ((1,), 0, [(slice(None, None),)]),
+        ((1, 2), 0, [(slice(None, None), slice(None, None))]),
+        (
+            (2, 4),
+            1,
+            [(0, slice(None, None)), (1, slice(None, None))],
+        ),
+    ],
+)
+def test_axis_ndindex(shape, axis, expected):
+    assert list(axis_ndindex(shape, axis)) == expected
 
 
 @pytest.mark.parametrize(
--- a/array_api_tests/meta/test_utils.py
+++ b/array_api_tests/meta/test_utils.py
@@ -1,7 +1,8 @@
 import pytest
 
-from ..test_signatures import extension_module
 from ..test_creation_functions import frange
+from ..test_signatures import extension_module
+from ..test_statistical_functions import axes_ndindex
 
 
 def test_extension_module_is_extension():
@@ -24,3 +25,21 @@ def test_extension_func_is_not_extension():
 def test_frange(r, size, elements):
     assert len(r) == size
     assert list(r) == elements
+
+
+@pytest.mark.parametrize(
+    "shape, axes, expected",
+    [
+        ((), (), [((),)]),
+        (
+            (2, 2),
+            (0,),
+            [
+                ((0, 0), (1, 0)),
+                ((0, 1), (1, 1)),
+            ],
+        ),
+    ],
+)
+def test_axes_ndindex(shape, axes, expected):
+    assert list(axes_ndindex(shape, axes)) == expected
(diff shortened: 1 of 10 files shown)


honno:
Just noting that I'm writing more in-depth tests for manipulation functions now.

honno:
> Just noting that I'm writing more in-depth tests for manipulation functions now.

Annnd statistical functions.

array_api_tests/test_statistical_functions.py
```diff
@@ -0,0 +1,365 @@
...
+    kw = data.draw(hh.kwargs(axis=axes(x.ndim), keepdims=st.booleans()), label="kw")
+
+    out = xp.mean(x, **kw)
+
+    ph.assert_dtype("mean", x.dtype, out.dtype)
+    _axes = normalise_axis(kw.get("axis", None), x.ndim)
+    assert_keepdimable_shape(
+        "mean", out.shape, x.shape, _axes, kw.get("keepdims", False), **kw
+    )
+    for indices, out_idx in zip(axes_ndindex(x.shape, _axes), ah.ndindex(out.shape)):
```
asmeurer, line 176: You should be able to use the _test_stacks helper in the linalg tests for this sort of thing. Except I'm currently in the process of updating it to handle stacks along arbitrary axes, so you'll need to wait for me to finish doing that first before you can.

honno: This works fine as-is. I'd rather refactor after the fact myself.

array_api_tests/meta/test_utils.py
```diff
@@ -24,3 +27,45 @@ def test_extension_func_is_not_extension():
...
+            (2, 2),
+            (0,),
+            [
+                ((0, 0), (1, 0)),
+                ((0, 1), (1, 1)),
+            ],
+        ),
+    ],
+)
+def test_axes_ndindex(shape, axes, expected):
```
honno, line 70: - [ ] redundant test now, will remove it

array_api_tests/test_statistical_functions.py
```diff
@@ -0,0 +1,365 @@
...
+        mean = float(out[out_idx])
+        elements = []
+        for idx in indices:
+            s = float(x[idx])
+            elements.append(s)
+        expected = sum(elements) / len(elements)
```
asmeurer, line 182: Does this work? I'm a little surprised if the result mean is the same as the naively computed mean. An array library may use an algorithm which is more numerically stable than this (e.g., it could [avoid cancellation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) in the sum, or by promoting to a higher dtype for intermediate calculations), or even less numerically stable depending on how it is implemented. So I don't think we can test exact values here.

honno: Haha yeah, it doesn't quite work yeah. I'm not testing exact values already (my `asset_equals()` uses `isclose()`). I will have to figure out a better/much more lenient method.

honno: I've just increased the rel and abs tol, seems fine for NumPy at least. It's pretty lenient now.

array_api_tests/test_manipulation_functions.py
```diff
@@ -0,0 +1,369 @@
...
+                        out[out_idx],
+                        **kw,
+                    )
+
+
+@given(
+    x=xps.arrays(dtype=xps.scalar_dtypes(), shape=shared_shapes()),
+    axis=shared_shapes().flatmap(lambda s: st.integers(-len(s) - 1, len(s))),
+)
+def test_expand_dims(x, axis):
```
asmeurer, line 148: TODO here: the spec requires the function to raise IndexError for invalid axis.

array_api_tests/test_manipulation_functions.py
```diff
@@ -0,0 +1,369 @@
...
+    )
+
+
+@given(
+    x=xps.arrays(
+        dtype=xps.scalar_dtypes(), shape=hh.shapes(min_side=1).filter(lambda s: 1 in s)
+    ),
+    data=st.data(),
+)
+def test_squeeze(x, data):
```
asmeurer, line 170: Let's keep the test functions in alphabetical order in the file.

asmeurer: This function also has to raise a specific error for invalid axes. Probably just generating all possible axes instead of only the valid ones can prevent the need for `data` here.

array_api_tests/test_manipulation_functions.py
```diff
@@ -0,0 +1,369 @@
...
+    # TODO: generate valid negative axis (which keep uniqueness)
```
asmeurer, line 171: I think you can use the `unique_by` flag here, like `unique_by=lambda i: i if i >= 0 else i + ndim + 1` along with `integers(-ndim, ndim-1)`.

honno:
Just nothing that there are some new utils that could be moved into the helper modules, but I'd rather finishing implementing the rest of the tests first so as to not prematurely refactor.

PR #43, opened 2021-11-17
Monkey patch Hypothesis to work around `st.floats()` bug

### LINKED ISSUE #42: Subnormals make testing flush-to-zero libs/builds impractical
Running the test suite against `cupy.array_api` I'm seeing a lot of the following error:
```
self = <hypothesis.extra.array_api.ArrayStrategy object at 0x7fd585346f40>, val = 5.877471754111438e-39, val_0d = Array(0., dtype=float32)
strategy = floats(width=32)

    def check_set_value(self, val, val_0d, strategy):
        finite = self.builtin is bool or self.xp.isfinite(val_0d)
        if finite and self.builtin(val_0d) != val:
                f"Generated array element {val!r} from strategy {strategy} "
                f"cannot be represented with dtype {self.dtype}. "
                f"Array module {self.xp.__name__} instead "
                f"represents the element as {val_0d}. "
                "Consider using a more precise elements strategy, "
                "for example passing the width argument to floats()."
            )
E           hypothesis.errors.InvalidArgument: Generated array element 5.877471754111438e-39 from strategy floats(width=32) cannot be represented with dtype float32. Array module cupy.array_api instead represents the element as 0.0. Consider using a more precise elements strategy, for example passing the width argument to floats().
```
I think the comparison `self.builtin(val_0d) != val` should be changed to `isclose(val_0d, val)`.

honno:
Resolves #42 by monkey patching the relevant `hypothesis.extra.array_api` internals to work around HypothesisWorks/hypothesis#3153. This is done by filtering the original `from_dtype()` strategy to throw away the bad floats inbetween the smallest representable non-zero number and zero.

### DIFF
--- a/array_api_tests/__init__.py
+++ b/array_api_tests/__init__.py
@@ -1,10 +1,27 @@
-from hypothesis.extra.array_api import make_strategies_namespace
+from hypothesis.extra import array_api
 
+from . import dtype_helpers as dh
 from ._array_module import mod as _xp
 
+__all__ = ["xps"]
 
-xps = make_strategies_namespace(_xp)
+
+# For now we monkey patch the internal _from_dtype() method to work around a bug
+# in st.floats() - see https://github.com/HypothesisWorks/hypothesis/issues/3153
 
 
-del _xp
-del make_strategies_namespace
+broken_from_dtype = array_api._from_dtype
+
+
+def _from_dtype(xp, dtype, **kwargs):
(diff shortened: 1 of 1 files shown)


asmeurer:
Looks fine, although if this is going to be fixed upstream soon it's not necessary to work around it here.

honno:
Closing as #45 achieved the goals of this PR.

PR #47, opened 2021-12-08
Monkey patch `st.floats()` to always disable subnormals

honno:
Now subnormals are out-of-scope in data-apis/array-api#341, this PR disables them completely by monkey-patching the underlying `st.floats()` strategy used in `xps.from_dtype()` and `xps.arrays()`. This is a bit hacky, but requires no change in how we use strategies i.e. no passing `allow_subnormal` or wrapping `xps.arrays`.

Tested with both NumPy and CuPy.

### DIFF
--- a/array_api_tests/__init__.py
+++ b/array_api_tests/__init__.py
@@ -1,10 +1,24 @@
+from functools import wraps
+
+from hypothesis import strategies as st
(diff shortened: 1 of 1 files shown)


honno:
Merging as it's a small change that's nice to get in ASAP.

PR #48, opened 2021-12-10
Test everything

honno:
This PR should finishing testing every function in the spec :tada: ~Starts from #39. Currently just implements searching tests.~ Covers everything from the now-closed #39, and finishes out the rest.

~I'd rather do this all in one go as 1) I'd rather not further chain PRs 2) I'm gunna want to refactor some utils.~


honno:
Also fixes #50

honno:
@asmeurer This PR makes significant changes to the manip and stat functions of #39, so I've closed that as it doesn't make sense to review such out-dated and somewhat erroneous code.

I'm also now ready for review/happy for merging.

asmeurer:
Yes, it would be OK with me if you do more self merging. I'm not able to review such large PRs in detail in a reasonable amount of time.

PR #52, opened 2021-12-17
Change scope of `test_constants.py`

### LINKED ISSUE #14: Test explicitly that constants are Python scalars

honno:
This PR:
- Removes dependencies on most `xp` functions, as they're not interesting IMO so unnecessarily increase dependencies
  - Note these tests were written before data-apis/array-api#169 where such use of arrays was necessary
  - Opted to use pure Python equalities instead
- Instance checks the constants should work like floats i.e. supports `__float__` 
  - `float` check would hurt duck typed implementations, see https://github.com/data-apis/array-api/issues/154#issuecomment-833746576
  - [`numerary`](https://posita.github.io/numerary/0.1/) stuff doesn't look necessary here (interesting for other parts of the suite tho)
- Adds interesting equality checks that @pmeier wrote for #15... useful if we're not assuming `float`

Clarifying we don't need constants to be specifically `float` should close #14 and #15.

### DIFF
--- a/array_api_tests/test_constants.py
+++ b/array_api_tests/test_constants.py
@@ -19,7 +19,7 @@ def assert_0d_float(name: str, x: Array):
 
 
 @pytest.mark.parametrize("name, n", [("e", math.e), ("pi", math.pi)])
-def test_irrational(name, n):
+def test_irrational_numbers(name, n):
     assert hasattr(xp, name)
     c = getattr(xp, name)
     assert_scalar_float(name, c)
--- a/array_api_tests/test_constants.py
+++ b/array_api_tests/test_constants.py
@@ -1,34 +1,43 @@
-from ._array_module import (e, inf, nan, pi, equal, isnan, abs, full,
-                            float64, less, isinf, greater, all)
-from .array_helpers import one
+import math
 
-def test_e():
-    # Check that e acts as a scalar
-    E = full((1,), e, dtype=float64)
+import pytest
 
-    # We don't require any accuracy. This is just a smoke test to check that
-    # 'e' is actually the constant e.
-    assert all(less(abs(E - 2.71), one((1,), dtype=float64))), "e is not the constant e"
+from . import dtype_helpers as dh
+from ._array_module import mod as xp
+from .typing import Array
 
-def test_pi():
-    # Check that pi acts as a scalar
-    PI = full((1,), pi, dtype=float64)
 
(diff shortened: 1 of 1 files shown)


array_api_tests/test_constants.py
```diff
@@ -1,34 +1,51 @@
...
+
+
+def assert_0d_float(name: str, x: Array):
+    assert dh.is_float_dtype(
+        x.dtype
+    ), f"xp.asarray(xp.{name})={x!r}, but should have float dtype"
+
+
+@pytest.mark.parametrize("name, n", [("e", math.e), ("pi", math.pi)])
+def test_irrational(name, n):
```
asmeurer, line 22: This test is actually testing that n is fractional, not irrational (floating point numbers aren't strictly "irrational", and it's impossible to test if they "represent" one).

array_api_tests/test_constants.py
```diff
@@ -1,34 +1,51 @@
...
+    assert isinstance(c, SupportsFloat), f"{name}={c!r} does not look like a float"
+
+
+def assert_0d_float(name: str, x: Array):
+    assert dh.is_float_dtype(
+        x.dtype
+    ), f"xp.asarray(xp.{name})={x!r}, but should have float dtype"
+
+
+@pytest.mark.parametrize("name, n", [("e", math.e), ("pi", math.pi)])
```
asmeurer, line 21: Shouldn't this be xp.e and xp.pi? Although why not just parameterize `name` and lookup `getattr(xp, name)` in the test?

array_api_tests/test_constants.py
asmeurer: I guess the "irrational" here means it's testing the irrational numbers. A bit confusing. Maybe `test_irrational_numbers` would be clearer.

asmeurer:
Just a few comments. Looks good otheriwse.

array_api_tests/test_constants.py
honno: Ah no, coz I'm using Python's `math` just to get the rough approximations (i.e. to test the `xp` constants are at least in-between the floor and ceil of the respective `math` constants).

PR #55, opened 2022-01-04
Better undefined dtype behaviour

honno:
This PR fixes some issues with undefined dtypes not being filtered from custom Hypothesis strategies, and additionally skips test cases which are explicitly passed undefined dtypes (i.e. for type promotion). cc @pmeier

### DIFF
--- a/xptests/hypothesis_helpers.py
+++ b/xptests/hypothesis_helpers.py
@@ -67,7 +67,13 @@ def _dtypes_sorter(dtype_pair: Tuple[DataType, DataType]):
         key += 1
     return key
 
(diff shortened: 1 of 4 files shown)


pmeier:
This does not fix the issue that I faced in

https://github.com/data-apis/array-api-tests/blob/e4ce31e2ee50af5b16e225bdd2cddda4507fe1a5/xptests/hypothesis_helpers.py#L53

The problem arises if we hit 

https://github.com/data-apis/array-api-tests/blob/e4ce31e2ee50af5b16e225bdd2cddda4507fe1a5/xptests/hypothesis_helpers.py#L58-L59

with an undefined dtype, since `_sorted_dtypes` only contains valid dtypes.

PR #58, opened 2022-01-05
Filter zero division examples for `test_remainder`

honno:
For integers, zero division is out-of-scope. cc @pmeier


honno:
There's another bug with the `not_nan` mask I see, must of been there for awhile. It also didn't check if the out array had a float dtype. I think more can be done to test the output of remainder, but for now I'll leave it all out so at least we have a smoke test.

PR #62, opened 2022-01-06
Update divide and pow input dtypes

### LINKED ISSUE #60: Update dtype strategies for elementwise/operators tests
Many of the elementwise/operator tests were written a while back, and there has been a lot of updates since in regards to what dtypes are allowed. This means we currently don't test all possible inputs!

Bug found today by @IvanYashchuk was `numpy.array_api.pow()` not working for integer arrays, probably because of the test suite was not updated with the spec.

So I'll need to review all the test methods to identify which aren't using all valid dtypes, and subsequently update them.

honno:
Resolves #60. Fortunately it was just these two that weren't updated.

Found that `numpy.array_api` doesn't accept int arrays for divide (as well as pow) numpy/numpy#20752


honno:
Didn't skip the special case and out-of-scope areas for pow, oops. Also divide turns out to be tricky data-apis/array-api#361

PR #66, opened 2022-01-07
Skip testing negative integer exponents for pow related functions

honno:
This is an out-of-scope scenario and thus should not be tested. I introduced testing numeric dtypes in #62 but did not think to do this at the time.

### DIFF
--- a/xptests/test_operators_and_elementwise_functions.py
+++ b/xptests/test_operators_and_elementwise_functions.py
@@ -1267,6 +1267,12 @@ def test_pow(
 ):
     left = data.draw(left_strat, label=left_sym)
     right = data.draw(right_strat, label=right_sym)
+    if right_is_scalar:
+        if isinstance(right, int):
+            assume(right >= 0)
+    else:
+        if dh.is_int_dtype(right.dtype):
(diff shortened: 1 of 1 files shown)


xptests/test_operators_and_elementwise_functions.py
```diff
@@ -1267,6 +1267,12 @@ def test_pow(
 ):
     left = data.draw(left_strat, label=left_sym)
     right = data.draw(right_strat, label=right_sym)
+    if right_is_scalar:
+        if isinstance(right, int):
+            assume(right >= 0)
+    else:
+        if dh.is_int_dtype(right.dtype):
+            assume(xp.all(right >= 0))
```
asmeurer, line 1275: I guess if hypothesis doesn't complain then we shouldn't worry about it, but wouldn't this filter a large fraction of examples?

honno: Yeah, Hypothesis hasn't complained yet here heh. `assume` is very useful as otherwise I'd need some mechanism to inject a test-specific strategy, but I'll mull over on if I can figure out a nicer (more performative!) solution here.

PR #67, opened 2022-01-07
Rename "xptests" back to "array_api_tests"

asmeurer:
See https://github.com/data-apis/array-api-tests/pull/53.

In the future, we should discuss changes like this first.


array_api_tests/test_sorting_functions.py
```diff
@@ -5,7 +5,7 @@
 from hypothesis import strategies as st
 from hypothesis.control import assume
 
-from xptests.typing import Scalar, ScalarType, Shape
+from .typing import Scalar, ScalarType, Shape
```
asmeurer, line 8: An aside: we need to avoid absolute imports like this, so that the module can easily be vendored. Maybe we should add a simple meta-test for this.

honno: Yep this was just an unintentional mistake

PR #73, opened 2022-01-12
Add a better mutually_broadcastable_shapes strategy

asmeurer:
This is based on the work from
https://github.com/Quansight-Labs/ndindex/pull/126. The strategy that comes
with hypothesis has default parameters that makes it not generate interesting
examples by default. In particular, it will never generate a shape with more
than 2 dimensions and side length more than 3.

This new strategy generates shapes that are based on a base shape drawn from
shapes(), so more complex shapes are possible. This is preferable to doing so
by modifying the parameters to mutually_broadcastable_shapes() because it's
difficult to find parameters that give interesting examples but don't require
too much filtering for total size.

This commit also changes it so that mutually_broadcastable_shapes generates
shapes with minimum side length 0 by default.

For example, before this commit:

```py
>>> from array_api_tests.hypothesis_helpers import mutually_broadcastable_shapes
>>> for i in range(10):
...     print(mutually_broadcastable_shapes(2).example())
((3,), (3, 1))
((), (3,))
((), (2,))
((1,), ())
((3,), ())
((), (1,))
((1,), (1, 1))
((2,), ())
((), (1,))
((3,), ())
```

After this commit:

```py
>>> for i in range(10):
...     print(mutually_broadcastable_shapes(2).example())
((), (1, 1))
((0, 1), (1, 1, 1, 1))
((1, 3, 1, 1), (3, 1, 4))
((), ())
((1, 5, 1), (1, 5, 1))
((1, 1), (2, 2))
((1,), (2, 0, 1))
((), (1, 1, 1, 2))
((), ())
((), (1, 4, 1))
```

The generation with this new strategy is somewhat slower than the old one
(about 2x slower), but the generation speed is not slow enough that it should
matter too much.

The shrinking can also be imperfect with this strategy. For example, I've seen
it shrink to ((0, 1), (0, 2)) for the currently failing in-place tests where
it could shrink to just ((1,), (2,)). I also saw the pre-existing strategy
have the same problem, however. One should just be aware that test failures
involving shapes from this strategy might not actually be shrunk minimally.

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -4,7 +4,7 @@ from math import sqrt
 from operator import mul
 from typing import Any, List, NamedTuple, Optional, Sequence, Tuple, Union
 
-from hypothesis import assume
+from hypothesis import assume, note
 from hypothesis.strategies import (SearchStrategy, booleans, composite, floats,
                                    integers, just, lists, none, one_of,
                                    sampled_from, shared)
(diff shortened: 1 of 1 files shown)


array_api_tests/hypothesis_helpers.py
```diff
@@ -164,16 +164,41 @@ def matrix_shapes(draw, stack_shapes=shapes()):
...
+    # be (). We are only using it here to trick mutually_broadcastable_shapes
+    # into giving more interesting examples.
+    final_result_shape = broadcast_shapes(*input_shapes)
+
+    # The broadcast compatible shapes can be bigger than the base shape. This
+    # is already somewhat limited by the mutually_broadcastable_shapes
+    # defaults, and pretty unlikely, but we filter again here just to be safe.
+    if not prod([i for i in final_result_shape if i]) < SQRT_MAX_ARRAY_SIZE:
+        note(f"Filtering in mutually_broadcastable_shapes {result_shape}")
+        assume(False)
```
honno, line 195:
```suggestion
    assume(prod(i for i in final_result_shape if i) < SQRT_MAX_ARRAY_SIZE)
```
Also we should avoid using notes like this except when debugging, as it creates unnecessary noise for test failures.

array_api_tests/hypothesis_helpers.py
```diff
@@ -164,16 +164,41 @@ def matrix_shapes(draw, stack_shapes=shapes()):
...
+    # https://github.com/HypothesisWorks/hypothesis/issues/3170). It's very
+    # difficult to get it to do so by tweaking the max_* parameters, because
+    # making them too big leads to generating too large shapes and filtering
+    # too much. So instead, we trick it into generating more interesting
+    # examples by telling it to create shapes that broadcast against some base
+    # shape.
+    if 'base_shape' not in kwargs:
+        base_shape = draw(shapes())
+        kwargs.setdefault('base_shape', base_shape)
+    kwargs.setdefault('min_side', 0)
```
honno, line 180:
```suggestion
    kwargs.setdefault('min_side', 0)
    if 'base_shape' not in kwargs:
        shape_kw = {}
        for k in ['min_dims', 'max_dims', 'min_side', 'max_side']:
            if k in kwargs.keys():
                shape_kw[k] = kwargs[k]
        base_shape = draw(shapes(**shape_kw))
        kwargs['base_shape'] = base_shape
```

Carry over kwargs when generating base shape, avoiding erroneous use of `xps.mutually_broadcastable_shapes()` below.

honno (changes requested):
Generating base shapes is a nice idea. Carrying over kwargs to `base_shape` is blocking (see review comment), otherwise this LGTM.

array_api_tests/hypothesis_helpers.py
asmeurer: Ah, it hadn't occurred to me that they both take the same keyword arguments.

honno:
We can merge this, but it's worth just checking if changing the defaulting does what you want—please try out #75. This would be preferable for minimisation behaviour, as well as be less of a maintenance burden.

asmeurer:
The simpler change https://github.com/data-apis/array-api-tests/pull/75 should work for now. If we ever decide to modify `shapes` from the default strategy, this may end up being a better approach. Ultimately, I'd prefer if hypothesis allowed filtering both shapes and mutually_broadcastable_shapes by total size in a more direct way.

PR #75, opened 2022-01-14
Change `hh.mutually_broadcastable_shapes()` defaults to generate more interesting shapes

honno:
Changes the defaulting of `max_dims` and `max_side` so that more interesting shapes get generated. Also defaults `min_side` to `0`.

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -165,10 +165,27 @@ finite_matrices = xps.arrays(dtype=xps.floating_dtypes(),
                                            allow_infinity=False))
 
 def mutually_broadcastable_shapes(
-    num_shapes: int, **kw
+    num_shapes: int,
+    *,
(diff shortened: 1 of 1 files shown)


asmeurer:
I guess this doesn't end up being that different for us because the hypothesis `shapes` strategy is also based on limiting the side length, whereas in ndindex I just use `tuples(integer())` and filter across the total size.

honno:
> I guess this doesn't end up being that different for us because the hypothesis `shapes` strategy is also based on limiting the side length, whereas in ndindex I just use `tuples(integer())` and filter across the total size.

Ah so you're saying #73 is limited as-is? Does this PR fair well? Maybe can bump up the defaults a bit more too.

My impression is that the defaults from Hypothesis might be too low, but I also think relying on `st.integers().filter(<below max size>)` is overkill (i.e. seemingly won't be finding interesting bugs for the big performance cost when making/manipulating arrays).

PR #86, opened 2022-01-25
Use ndindex.iter_indices in _test_stacks in the linalg tests

asmeurer:
This adds ndindex >= 1.6 as a dependency of the test suite.

I'm also going to try to make the tests pass. 

I want to update some of the existing tests to use this as well, but I am going to do that in a second PR because it shouldn't block a release.


asmeurer:
This should be ready to go. I'll make my further changes in another PR so that we can do a release now.

honno:
LGTM. Great to see a `ndindex` release!

asmeurer:
Yeah. Sorry it took so long. Let me know if you run across any other complex indexing related things that you think ndindex could help with.

PR #89, opened 2022-01-26
Increase coverage and other niceties for the operator/elementwise tests

### LINKED ISSUE #71: Pretty print indices
Right now indices are usually printed as tuples e.g. `out[(1, 3)]`, but it'd be nice if they were reduced e.g. `out[1, 3]`. I should review `pytest_helpers.py` at least to do this with a new util. This goes for ellipsis and slices too.

### LINKED ISSUE #84: Test shape broadcasting for _all_ elementwise/operator tests
Coverage of broadcasting is spotty here, so I need to review each test and see what needs to be done.

honno:
This PR:
- Utilities `ndindex.iter_indices()` to greatly improve values testing
  - This is used for _every_ test via the newly introduced `*_assert_against_refimpl()` utils
- Implement values testing for **all** remaining op/elwise tests
- Test scalar scenarios for binary parametrized tests  (i.e. the ones that cover both operators and elementwise functions)
- Greatly improves error output (no more million random variables popping up when there's an error)

Additionally:
- Resolves #71 with `sh.fmt_idx()`
- Resolves #84

### DIFF
--- a/array_api_tests/test_operators_and_elementwise_functions.py
+++ b/array_api_tests/test_operators_and_elementwise_functions.py
@@ -66,14 +66,12 @@ def unary_assert_against_refimpl(
     res: Array,
     refimpl: Callable[[Scalar], Scalar],
     expr_template: str,
-    in_stype: Optional[ScalarType] = None,
     res_stype: Optional[ScalarType] = None,
     filter_: Callable[[Scalar], bool] = math.isfinite,
 ):
     if in_.shape != res.shape:
         raise ValueError(f"{res.shape=}, but should be {in_.shape=}")
-    if in_stype is None:
-        in_stype = dh.get_scalar_type(in_.dtype)
+    in_stype = dh.get_scalar_type(in_.dtype)
     if res_stype is None:
         res_stype = in_stype
     m, M = dh.dtype_ranges.get(res.dtype, (None, None))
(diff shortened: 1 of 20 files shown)


honno:
@asmeurer Not quite ready yet, but you might wanna skim over to see if you can identify any issues with the "reference implementation" approach I'm using here... it's a big change :sweat_smile:

Notably this does still pass-but-filter special cases (and without using boolean masks). I wrote a high-level note here: https://github.com/data-apis/array-api-tests/blob/9d1f4da9ad508a00313a56574387b6178e8ca7cc/array_api_tests/test_operators_and_elementwise_functions.py#L45-L65

asmeurer:
Let's make sure we document that we're doing this and that any errors should usually mean the function is wrong, but it could also mean our tolerances need to be loosened.

asmeurer:
My biggest worry here is float32, since the reference implementation will operate on float64. And if we ever add a 16-bit float, that will be even worse. You might want to test this on something like cupy just to see if it works on an accelerator.

array_api_tests/test_operators_and_elementwise_functions.py
```diff
@@ -25,67 +15,215 @@
...
+# By default, results are casted to scalars the same way that the inputs are.
+# You can specify a cast via `res_stype, i.e. when a function accepts numerical
+# inputs but returns boolean arrays.
+#
+# By default, floating-point functions/methods are loosely asserted against. Use
+# `strict_check=True` when they should be strictly asserted against, i.e.
+# when a function should return intergrals.
+
+
+def isclose(a: float, b: float, rel_tol: float = 0.25, abs_tol: float = 1) -> bool:
```
asmeurer, line 68: I guess since we are testing one float at a time, just using math.isclose here is fine. In general though, we will want an isclose, and corresponding assert_approximately_equal method that only use the array API methods so that they can be called on arrays (I will need this for the linalg tests).

I don't *think* it will be an issue, but if we do run into issues with float32 and the fact that Python `float` is 64-bit only, we may need to make this use array operations too. That would also give us more control over the exact formula used for "closeness", though I doubt that will be an issue either given our very loose tolerances.

honno: > In general though, we will want an isclose, and corresponding assert_approximately_equal method that only use the array API methods so that they can be called on arrays (I will need this for the linalg tests).

The `*_assert_against_refimpl` utils do this along an array/arrays, only using the Array API endpoints of indexing and scalar casting.

> I don't think it will be an issue, but if we do run into issues with float32 and the fact that Python float is 64-bit only, we may need to make this use array operations too.

Ah I see, good to keep in mind.

honno:
Happy with this PR now.

PR #101, opened 2022-02-18
More improvements to test_linalg

asmeurer:
So far just vector_norm, but more are incoming.

Note that NumPy will fail this test without https://github.com/numpy/numpy/pull/21084.

I also realized that the linalg tests should be making better use of the helper functions. I've done so for this test, but still need to update the other tests.


array_api_tests/test_linalg.py
```diff
@@ -645,11 +644,52 @@ def test_vecdot(dtypes, shape, data):
...
+                   floats(1, max_ord, allow_nan=False,
+                          allow_infinity=False),
+                   just(0),
+                   just(-1),
+                   just(-2),
+                   just(float('-inf')),
+                   integers(-max_ord, -1),
+                   floats(-max_ord, -1, allow_nan=False,
+                          allow_infinity=False),
+               )), label="kw")
```
honno, line 675: How about using:
```python
ord=one_of(
    integers(-max_ord, max_ord),
    floats(-max_ord, max_ord),
    sampled_from([float("inf"), float("-inf")]),
)
```
Note setting both bounds in `st.floats()` will disable generation of NaNs and infinities. Were you concerned about minimisation behaviour?

honno (commented):
Looks good.

> Note that NumPy will fail this test without [numpy/numpy#21084](https://github.com/numpy/numpy/pull/21084).

Ah, we should skip this in the workflow anywho because we're pinning against NumPy releases now.

array_api_tests/test_linalg.py
honno, line 675: How about something like:
```suggestion
               ord=one_of(
                   sampled_from([2, 1, 0, -1, -2, float("inf"), float("-inf")],
                   integers(-max_ord, max_ord),
                   floats(-max_ord, max_ord),
               )), label="kw")
```
Note setting bounds of `st.floats()` will disable generation of nans and infs anyway.

asmeurer: I wasn't sure about the nans thing. I wanted to make sure it was more likely to pick the integers from -2 to 2 since those are the important orders.

honno: Yeah, all the samples should be prioritised anywho due to their order in `one_of()` and the small sample size.

asmeurer:
@honno see what you think of the allclose helpers I've added here. They can likely use some improvement, but they should work for now.

array_api_tests/array_helpers.py
```diff
@@ -146,6 +149,43 @@ def exactly_equal(x, y):
...
+        b = y[j]
+        if not (math.isfinite(a) and math.isfinite(b)):
+            # TODO: If a and b are both infinite, require the same type of infinity
+            continue
+        close = math.isclose(a, b, rel_tol=rel_tol, abs_tol=abs_tol)
+        if not close:
+            if return_indices:
+                return (False, (i, j))
+            return False
+    return True
```
honno, line 172: Yep looks good, this is really how we have to do it if we can't use masking. Maybe move this to `test_linalg` for now, as ideally what we'd be doing is standardising these utils across function groups then, and scoping these utils for each function group makes it easier to distinguish where they're used. (standardising seems quite doable, I just need to get round to it)

asmeurer: One problem here is represents a pretty significant performance hit. Before this commit (with the exact equality test), the linalg tests take 15 seconds on my computer. After, they take 44 seconds. Performance isn't our top priority, but maybe we should try array operations and only fallback to this when there are nonfinite values.

asmeurer:
It looks like it isn't as straightforward as this. CuPy has several failures:

```
_________________________________________________________________________________________________________ test_eigh __________________________________________________________________________________________________________

    @pytest.mark.xp_extension('linalg')
>   @given(x=symmetric_matrices(finite=True))

array_api_tests/test_linalg.py:239: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
array_api_tests/test_linalg.py:256: in test_eigh
    _test_stacks(lambda x: linalg.eigh(x).eigenvectors, x,
array_api_tests/test_linalg.py:90: in _test_stacks
    assert_equal(res_stack, decomp_res_stack)
array_api_tests/test_linalg.py:49: in assert_equal
    assert_allclose(x, y)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = Array([[-0.40824828,  0.70710677,  0.57735026],
       [-0.40824828, -0.70710677,  0.57735026],
       [ 0.81649655,  0.        ,  0.57735026]], dtype=float32)
y = Array([[ 0.8164966 ,  0.        , -0.57735026],
       [-0.4082483 , -0.70710677, -0.5773504 ],
       [-0.40824828,  0.7071068 , -0.5773503 ]], dtype=float32), rel_tol = 0.25, abs_tol = 1

    def assert_allclose(x, y, rel_tol=0.25, abs_tol=1):
        """
        Test that x and y are approximately equal to each other.
    
        Also asserts that x and y have the same shape and dtype.
        """
        assert x.shape == y.shape, f"The input arrays do not have the same shapes ({x.shape} != {y.shape})"
    
        assert x.dtype == y.dtype, f"The input arrays do not have the same dtype ({x.dtype} != {y.dtype})"
    
        c = allclose(x, y, rel_tol=rel_tol, abs_tol=abs_tol, return_indices=True)
        if c is not True:
            _, (i, j) = c
>           raise AssertionError(f"The input arrays are not close with {rel_tol = } and {abs_tol = } at indices {i = } and {j = }")
E           AssertionError: The input arrays are not close with rel_tol = 0.25 and abs_tol = 1 at indices i = (0, 0) and j = (0, 0)

array_api_tests/array_helpers.py:187: AssertionError
--------------------------------------------------------------------------------------------------------- Hypothesis ---------------------------------------------------------------------------------------------------------
Falsifying example: test_eigh(
    x=Array([[[1., 1., 1.],
            [1., 1., 1.],
            [1., 1., 1.]]], dtype=float32),
)
___
```

Note that the sort order of eigenvalues is unspecified in the spec.

Also svd and svdvals fail with errors like

```
________________________________________________________________________________________________________ test_svdvals ________________________________________________________________________________________________________

    @pytest.mark.xp_extension('linalg')
>   @given(
        x=finite_matrices(),
    )

array_api_tests/test_linalg.py:562: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
array_api_tests/test_linalg.py:577: in test_svdvals
    _test_stacks(linalg.svdvals, x, dims=1, res=res)
array_api_tests/test_linalg.py:90: in _test_stacks
    assert_equal(res_stack, decomp_res_stack)
array_api_tests/test_linalg.py:49: in assert_equal
    assert_allclose(x, y)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = Array([34741248.,        0.], dtype=float32), y = Array([3.4741252e+07, 1.3871019e+00], dtype=float32), rel_tol = 0.25, abs_tol = 1

    def assert_allclose(x, y, rel_tol=0.25, abs_tol=1):
        """
        Test that x and y are approximately equal to each other.
    
        Also asserts that x and y have the same shape and dtype.
        """
        assert x.shape == y.shape, f"The input arrays do not have the same shapes ({x.shape} != {y.shape})"
    
        assert x.dtype == y.dtype, f"The input arrays do not have the same dtype ({x.dtype} != {y.dtype})"
    
        c = allclose(x, y, rel_tol=rel_tol, abs_tol=abs_tol, return_indices=True)
        if c is not True:
            _, (i, j) = c
>           raise AssertionError(f"The input arrays are not close with {rel_tol = } and {abs_tol = } at indices {i = } and {j = }")
E           AssertionError: The input arrays are not close with rel_tol = 0.25 and abs_tol = 1 at indices i = (1,) and j = (1,)

array_api_tests/array_helpers.py:187: AssertionError
```

which shows that using an absolute tolerance is not going to work. 

There are also some test failures which I think are due to bugs in cupy. This one I'm not sure about

```py
>>> import cupy.array_api
>>> x1 = cupy.array_api.asarray([51307], dtype=cupy.array_api.uint16)
>>> x2 = cupy.array_api.asarray([[[327]]], dtype=cupy.array_api.uint16)
>>> cupy.array_api.linalg.matmul(x1, x2)
Array([[172]], dtype=uint16)
>>> cupy.array_api.linalg.matmul(x1, x2[0, ...])
Array([173], dtype=uint16)
>>> x1
Array([51307], dtype=uint16)
>>> import numpy as np
>>> np.asarray([51307*327], dtype=np.uint16)
array([173], dtype=uint16)
```

CuPy is clearly doing something weird here, but should this overflow behavior be something that we should be avoiding in the test suite? If so then I guess we need to be more careful about how we generate integer array inputs to functions like matmul.

honno:
> should this overflow behavior be something that we should be avoiding in the test suite?

Ah interesting catch. Personally I think it's fine to be relative even with integer dtypes here, as it probably won't affect interop.

array_api_tests/test_linalg.py
```diff
@@ -46,7 +46,15 @@
...
-        assert_allclose(x, y)
+        # It's too difficult to do an approximately equal test here because
+        # different routines can give completely different answers, and even
+        # when it does work, the elementwise comparisons are too slow. So for
+        # floating-point dtypes only test the shape and dtypes.
+
+        # assert_allclose(x, y)
+
+        assert x.shape == y.shape, f"The input arrays do not have the same shapes ({x.shape} != {y.shape})"
+        assert x.dtype == y.dtype, f"The input arrays do not have the same dtype ({x.dtype} != {y.dtype})"
```
honno, line 57: Yep this seems reasonable.

asmeurer:
What needs to happen to get this merged? Note that vecdot will fail with NumPy until https://github.com/numpy/numpy/pull/21928 is merged.

asmeurer:
Also test_solve is commented out until I finish https://github.com/Quansight-Labs/ndindex/pull/131.

honno (commented):
Had another gander, LGTM! Just needs fixing the failing test on CI ofc, which looks like an issue with 0-sided arrays ugh.

honno:
@asmeurer if you have the time could you look into updating this? Otherwise if you don't have the bandwidth lemme know.

asmeurer:
Not sure if I'll have the bandwidth in the immediate future, but it is still on my list of things to do. If you want to take this over I can go over what needs to be done.

honno:
> Not sure if I'll have the bandwidth in the immediate future, but it is still on my list of things to do. If you want to take this over I can go over what needs to be done.

Yeah lets touch base after Thanksgiving week

asmeurer:
@honno I've been cleaning this up. It now depends on an upstream ndindex fix which I need to release https://github.com/Quansight-Labs/ndindex/pull/155. The tests are now passing on array-api-strict (with https://github.com/data-apis/array-api-strict/pull/10). I still need to test numpy, cupy, and torch. 

It would help if you could look this over and see if anything looks off. I'm worried some of the code conventions here might have fallen out of date with the rest of the codebase. I've done my best to update the things I noticed, but I probably missed something. 

Also note that I did update the tests here for https://github.com/data-apis/array-api/pull/740, even though it isn't merged or decided on yet.

honno (commented):
Awesome LGTM, sans ndindex fixes and trivial linting issue. As far as I can tell you've utilised all the relevant helpers introduced since the PR was opened, which is grand to see!

array_api_tests/array_helpers.py
asmeurer: Actually I'm just going to delete them. I'm not using them anymore, because testing float arrays from linalg functions like this turned out to be too difficult.

asmeurer:
I've released ndindex 1.8 with the changes needed here, and updated the pin in requirements.txt.

honno:
Can ignore flaky statistical tests (skipped in `master` now since #233). Looks good, just need resolution on the failing `test_cross`.

asmeurer:
The test failure is because numpy.array_api is wrong for `cross` (I never added broadcasting support from the 2022 spec). array-api-strict has already been fixed. We will need to figure out what we want to do going forward for CI tests, given that numpy.array_api is going to be removed and array-api-strict is in its own repo that has its own array-api-tests CI run. 

At any rate, I just remembered that I still wanted to check this with numpy, torch, and cupy. Although if you want to merge anyway, you can, and I'll just make any follow up fixes in another PR.

asmeurer:
Submitted upstream https://github.com/numpy/numpy/issues/25840. What's the easiest way to skip this test for now?

asmeurer:
There are some other test failures with NumPy:

- `trace` fails because of https://github.com/data-apis/array-api/issues/731. I propose we fix that after this is merged (I have factored out the sum/prod/trace logic into a separate function here)
- `test_solve` fails with

```
E           ValueError: solve: Input operand 1 does not have enough dimensions (has 1, gufunc core with signature (m,m),(m,n)->(m,n) requires 2)
E           Falsifying example: test_solve(
E               x1=array([], shape=(0, 0, 0), dtype=float32),
E               x2=array([], dtype=float32),
E           )

```

(need to investigate)

- test_matrix_rank fails with

```
    | Traceback (most recent call last):
    |   File "/Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_linalg.py", line 429, in test_matrix_rank
    |     linalg.matrix_rank(x, **kw)
    |   File "/Users/aaronmeurer/Documents/numpy/numpy/linalg/_linalg.py", line 2083, in matrix_rank
    |     return count_nonzero(S > tol, axis=-1)
    |                          ^^^^^^^
    | ValueError: operands could not be broadcast together with shapes (3,2) (3,3)
    | Falsifying example: test_matrix_rank(
    |     x=array([[[0., 0.],
    |             [0., 0.]],
    |
    |            [[0., 0.],
    |             [0., 0.]],
    |
    |            [[0., 0.],
    |             [0., 0.]]], dtype=float32),
    |     kw={'rtol': array([0., 0., 0.], dtype=float32)},
    | )
    +---------------- 2 ----------------
    | Traceback (most recent call last):
    |   File "/Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_linalg.py", line 429, in test_matrix_rank
    |     linalg.matrix_rank(x, **kw)
    |   File "/Users/aaronmeurer/Documents/numpy/numpy/linalg/_linalg.py", line 2080, in matrix_rank
    |     tol = S.max(axis=-1, keepdims=True) * rtol
    |           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
    | ValueError: operands could not be broadcast together with shapes (0,2,1) (0,2)
    | Falsifying example: test_matrix_rank(
    |     x=array([], shape=(0, 2, 1, 1), dtype=float32),
    |     kw={'rtol': array([], shape=(0, 2), dtype=float32)},
    | )
```

(also need to investigate)

- I get a health check error for filtering too much with `invertable_matrices` from `test_inv` sometimes.

honno:
> Submitted upstream [numpy/numpy#25840](https://github.com/numpy/numpy/issues/25840). What's the easiest way to skip this test for now? (EDIT: looks like `-k 'not cholesky'` works)

Looks like it was fixed already :tada:

> We will need to figure out what we want to do going forward for CI tests, given that numpy.array_api is going to be removed and array-api-strict is in its own repo that has its own array-api-tests CI run.

I wrote https://github.com/data-apis/array-api-tests/issues/235 to track this and other musings on our CI. Probably for this PR we're sticking with `numpy.array_api`.

> `trace` fails because of [Reconsider sum/prod/trace upcasting for floating-point dtypes array-api#731](https://github.com/data-apis/array-api/issues/731). I propose we fix that after this is merged (I have factored out the sum/prod/trace logic into a separate function here)

Yep agreed can fix after, wrote https://github.com/data-apis/array-api-tests/issues/234 for this.

---

Regarding `test_solve` and `test_matrix_rank` failures (+ `test_inv` flakiness), yeah if this week you can identify the issues and either fix or slap a `pytest.mark.skip` if non-trivial, otherwise I think it'd be nice to just merge this PR and look at later.

asmeurer:
The np solve issue in array-api-compat is fixed here https://github.com/data-apis/array-api-compat/pull/93. Unfortunately, I don't think we can fix cupy as it doesn't seem to actually support stacking in its solve().

asmeurer:
Here's the invertable_matrices strategy that is giving the health error for too much filtering: https://github.com/data-apis/array-api-tests/blob/4f83bb3ee9146c30cf997f1527588b7a8b8ee6db/array_api_tests/hypothesis_helpers.py#L306-L318

Is there a more direct way to generate arrays `d` s.t. `all(abs(d) > 0.5)`?

(of course, I should probably come up with some way to generate more interesting invertible matrices than just diagonal ones).

asmeurer:
Looks like the matrix_rank thing is a legitimate bug in the numpy implementation https://github.com/numpy/numpy/pull/25437/files#r1500043325.

asmeurer:
@honno let me know what you think of the latest commit as a way to test functions like `matmul` which should be in both the main and linalg namespace separately. Previously only the linalg one was tested, but I want to make sure that the compat library properly wraps things in both namespaces.

honno (approved):
> let me know what you think of the latest commit as a way to test functions like matmul which should be in both the main and linalg namespace separately. Previously only the linalg one was tested, but I want to make sure that the compat library properly wraps things in both namespaces.

Yeah looks good, nice catch :+1:

> Is there a more direct way to generate arrays `d` s.t. `all(abs(d) > 0.5)`?

Pushed a change which does this by passing a customer `elements` strategy.

Lets merge as-is and resolve everything else in separate PRs :tada:

Issue #40, opened 2021-11-10
Make special case tests only use relevant arrays


honno:
Right now we generate arrays irrelevant to test cases, e.g. arrays with no NaNs in a test case about NaN behaviour. A low-priority thing is to generate test cases which are relevant to the test case (either by `xps.arrays(..., elements={...})` or `xps.arrays(...).filter(...)`). Probably requires some significant reworking of `generate_stubs.py`.

To quote myself from #38:

> It seems like the only problem with `numpy.array_api` (when running 2000 examples) is in-place addition.
> 
> ```python
> >>> x1 = xp.asarray(-0.0)
> >>> x1
> Array(-0., dtype=float64)
> >>> x2 = xp.asarray(-0.0)
> >>> xp.add(x1, x2)
> Array(-0., dtype=float64)
> >>> x1 + x2
> Array(-0., dtype=float64)
> >>> x1 += x2
> >>> x1
> Array(0., dtype=float64)  # not negative 0, like it is with xp.add() and __add__()
> ```
> 
> Interestingly I had to run like 200 examples to get this error consistently. Something for the future would be to filter strategies from the `@given()` level so that only arrays with NaNs/infs/whateverproperty-we're-testing are generated, which would mitigate this problem.

Issue #94, opened 2022-02-06
Avoid using boolean array indexing in special case tests


tomwhite:
[Boolean array indexing](https://data-apis.org/array-api/latest/API_specification/indexing.html#boolean-array-indexing) is optional in the spec since the output shape is data dependent. Libraries like Dask create output arrays where the shape is unknown, which causes the special case tests to fail since they use masking in their implementation. It would be good if these tests could avoid using boolean array indexing so the special cases can be checked in Dask.

PR #104, opened 2022-02-28
Utilise RST conversion of spec, generate runtime special cases tests

honno:
Ok, this ones a biggie. It's not finished yet, but thought you'd like to take a look @asmeurer.

The first big thing is making use of the RST spec (data-apis/array-api#343). This is achieved by adding the spec repo `array-api` as a git submodule inside of our `array-api-tests/` folder, adding the signatures package to the Python path, and extracting all the function objects (or class in the case of the array object).

(I tried not adding the signatures package to the Python path, as it's rather unseemly and potentially problematic to users, but I struggled trying to parse the functions/methods with the nifty `importlib` utils due to relative imports.)

Next we have the special case tests.
- As we're resolving #94, I've replaced using boolean masks to filter away irrelevant array elements, with filter functions that use the scalar (i.e. `float`) of the respect array element. These are constructed in `parse_cond()`. Results are now checked against the scalar too, with a result-checker function constructed in `parse_result()`.
  - Much of your regex exists in these condition/check-result "factories".
- For the special cases in binary functions, I've generalised how we parse every condition in a special case
  - No more regexes for each arrangement e.g. `TWO_ARGS_EQUAL`, `TWO_ARGS_EQUAL__LESS`, etc. :tada:
- If we can't parse something, we raise a warning most times. This should tell us if `parse_cond()`/`parse_result()`/something else needs updating to cover any new special cases with unforeseen properties.
- Special cases are represented in dataclasses, and parametrized with generalised test methods, enabling runtime tests as opposed to having statically generated tests.
- I was trying running all special case tests at once per function, but will return to the one-case one-test pattern.
  - I'm hooking `assume()` to the generated examples, so Hypothesis will complain if we're not actually generating relevant examples, resolving #40

I still have a fair bit to do, but I'm pretty happy with the architecture now. I will probably touch the signature tests or module stubbing in a later PR.

Note right now there's a bug with the conditions giving false-negatives/postives, due to some pass-by-reference problems when I defined local functions in loops :sweat_smile:


asmeurer:
> (I tried not adding the signatures package to the Python path, as it's rather unseemly and potentially problematic to users, but I struggled trying to parse the functions/methods with the nifty importlib utils due to relative imports.)

This seems fine. If you're worried about name conflicts we could rename it to something more meaningful like `array_api_signatures`.

asmeurer:
This looks good so far. Some thoughts:

- I guess you eliminated the need for the `regex` module? 
- As long as we are cleaning up the regular expressions, it might be a good idea to replace numbered groups with named groups. 
- A seems like a lot of the code in test_special_cases could be simplified with functools.partial. Although I'm also wondering if so many levels of indirection are actually necessary in the first place. 
- I typically use UPPERCASE for compiled regex objects. I'm actually unsure how common this practice is, though. 
- We should remove the duplication in the function_stubs submodule. I think we can replace the whole thing with a single file using module-level `__getattr__`. Or we could just remove it entirely and require the use of `signatures` to access the stubs. 
- It might be a good idea to have some (really fast to run) meta/smoke tests that we can add to the main array-api repo CI to make sure the `signatures` submodule doesn't break on us. 
- It would be helpful to have more comments in the code explaining what is going on, and short docstrings for all the internal helper functions.

array_api_tests/stubs.py
```diff
@@ -0,0 +1,47 @@
...
+EXTENSIONS = ["linalg"]
+extension_to_funcs: Dict[str, List[FunctionType]] = {}
+for ext in EXTENSIONS:
+    mod = name_to_mod[ext]
+    objects = [getattr(mod, name) for name in mod.__all__]
+    assert all(isinstance(o, FunctionType) for o in objects)
+    extension_to_funcs[ext] = objects
+
+
+sys.path.remove(spec_abs_path)
```
asmeurer, line 47: Keeping sys.path clean probably isn't that important as people generally shouldn't be importing array_api_tests (and the import would fail without `ARRAY_API_TESTS_MODULE` set anyway), but if we want to be defensive here, this should go in a `finally` block.

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,646 @@
...
+
+
+def parse_inline_code(inline_code: str) -> float:
+    if m := r_code.match(inline_code):
+        return parse_value(m.group(1))
+    else:
+        raise ValueParseError(inline_code)
+
+
+r_not = re.compile("not (?:equal to )?(.+)")
```
asmeurer, line 151: It's good to always use make regular expressions raw strings, even if they don't have backslashes. You don't want to add a backslash escape later only to not realize the string isn't raw.

asmeurer:
> If we can't parse something, we raise a warning most times. This should tell us if parse_cond()/parse_result()/something else needs updating to cover any new special cases with unforeseen properties.

Just to be sure, this should already be the case with the existing `generate_stubs.py`, no? We definitely want to fail if we come across anything we don't recognize.

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,646 @@
...
+            assert in_sign == "" and value_array != input_array  # sanity check
+            expr = f"{in_sign}x{input_array}ᵢ == {value_sign}x{value_array}ᵢ"
+            if value_array == "1":
+                if value_sign != "-":
+
+                    def cond(i1: float, i2: float) -> bool:
+                        eq = make_eq(i1)
+                        return eq(i2)
+
+                else:
```
asmeurer, line 390: Might be a good idea to code these more defensively like

```py
if value_sign == '+':
    ...
elif value_sign == '-':
    ...
else:
    raise ValueParseError
```

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,646 @@
...
+    match = r_special_cases.search(docstring)
+    if match is None:
+        return []
+    lines = match.group(1).split("\n")[:-1]
+    cases = []
+    for line in lines:
+        if m := r_case.match(line):
+            case_str = m.group(1)
+        else:
+            warn(f"line not machine-readable: '{line}'")
```
asmeurer, line 537: We really want this to be an error, not a warning, I think.

asmeurer:
> It might be a good idea to have some (really fast to run) meta/smoke tests that we can add to the main array-api repo CI to make sure the signatures submodule doesn't break on us.

Actually, I just realized the (git) submodule breaks the ability to vendor array_api_tests. I think we can make vendoring possible by allowing copying `signatures` into `array_api_tests` as a (Python) submodule. Or we can just copy it manually and commit it, similar to what `generate_stubs.py` does. 

Either way, it makes sense to keep `.function_stubs` as the primary way to access the stubs.

array_api_tests/test_special_cases.py
asmeurer: And even if we didn't want this to be an error, I would avoid using `warnings`. pytest is just going to list warnings at the end, but modules like NumPy already emit dozens of warnings when running the tests, which can be ignored, and any legitimate warning would be completely lost.

honno:
>     * I guess you eliminated the need for the `regex` module?

Yep, I was initially using `?!` like you did but ended up with a solution that doesn't need it.

>     * As long as we are cleaning up the regular expressions, it might be a good idea to replace numbered groups with named groups.

I did not know this was a thing, very nifty! Will explore using them.

>     * A seems like a lot of the code in test_special_cases could be simplified with functools.partial. Although I'm also wondering if so many levels of indirection are actually necessary in the first place.

Ahh `partial` looks very handy, wasn't really aware of them before. And yeah generally there might be room to reduce the indirection.

>     * I typically use UPPERCASE for compiled regex objects. I'm actually unsure how common this practice is, though.

Hmm, I've seen uppercase elsewhere, and `re_*`/`r_*` too. I like `r_*` but not religious about it.

>     * We should remove the duplication in the function_stubs submodule. I think we can replace the whole thing with a single file using module-level `__getattr__`. Or we could just remove it entirely and require the use of `signatures` to access the stubs.

I'll have a quick look at replacing the submodule, but as I'm not tackling module stubbing/signature tests in this PR I won't prioritise this just yet. As you mention it, `__getattr__` seems useful for module stubbing too.

>     * It might be a good idea to have some (really fast to run) meta/smoke tests that we can add to the main array-api repo CI to make sure the `signatures` submodule doesn't break on us.

I do have a few hard, non-test assertions for all of this, but definitely something I can review.

> Actually, I just realized the (git) submodule breaks the ability to vendor array_api_tests. I think we can make vendoring possible by allowing copying `signatures` into `array_api_tests` as a (Python) submodule. Or we can just copy it manually and commit it, similar to what `generate_stubs.py` does.

Hmm, as I'm sure you'd agree it'd be ideal if we didn't hardcode things from a maintenance perspective. I guess the question is if it's too annoying to ask users to pull and deal with an `array-api` submodule? I'd rather err on the side of maintainability, and maybe change tact if folk complain.

>     * It would be helpful to have more comments in the code explaining what is going on, and short docstrings for all the internal helper functions.

Yep, once I stabilise things :sweat_smile:

array_api_tests/stubs.py
honno: Ah I see. Yeah will just remove this line.

array_api_tests/test_special_cases.py
honno: Hmm, I believe the regex match should guarantee that value_sign is only "", "-" or "+". A `raise` here does work well as a sanity check tho.

array_api_tests/test_special_cases.py
honno: I'd rather not raise an error here as it's not critical (unlike the assertions I make in `stubs.py`), so would unnecessarily halt use of the test suite for the myriad of reasons a user could have an out-of-sync `array-api` submodule.

I agree warnings are pretty awkward in practice, but I still think they're the only appropriate choice we have. I was actually using this in `setup.cfg` when developing all the parsing code paths:
```python
[tool:pytest]
filterwarnings =
    ignore::ImportWarning
    ignore::RuntimeWarning
    ignore::DeprecationWarning
```

asmeurer:
> Hmm, as I'm sure you'd agree it'd be ideal if we didn't hardcode things from a maintenance perspective. I guess the question is if it's too annoying to ask users to pull and deal with an array-api submodule? I'd rather err on the side of maintainability, and maybe change tact if folk complain.

I don't think we should require vendored users to use a submodule. If we don't commit the signature files, we need a way for users to do it, like a script that packages a standalone vendor module that people can use, meaning they have to run that each time they update. So it's a question of trading some maintenance upside for us with some usability for end-users. OTOH maybe this is just a reason why we should upload a package so that people who vendor can copy the from that.

honno:
> > Hmm, as I'm sure you'd agree it'd be ideal if we didn't hardcode things from a maintenance perspective. I guess the question is if it's too annoying to ask users to pull and deal with an array-api submodule? I'd rather err on the side of maintainability, and maybe change tact if folk complain.
> 
> I don't think we should require vendored users to use a submodule. If we don't commit the signature files, we need a way for users to do it, like a script that packages a standalone vendor module that people can use, meaning they have to run that each time they update. So it's a question of trading some maintenance upside for us with some usability for end-users. OTOH maybe this is just a reason why we should upload a package so that people who vendor can copy the from that.

Ok yeah I see. What do we mean when we say vendored though? Like the only practical use we're seeing right now is pulling the repo and running it in CI, where the submodule requirement is only a minor inconvenience. If a different use case arises then it may be appropriate to support it only then.

Anywho, I think if we hard copy `signatures`, I'd want to set up a `pre-commit` script that checks our own vendored `signatures` is synced with a version we specify somewhere. This PR is so big already that I'd want to do all that in another PR at once.

honno:
https://github.com/data-apis/array-api-tests/issues/40 is pretty tricky. It seems there's a fair few existing special case tests that don't generate that many relevant examples. so it could be apt to instead generate a single minimal reproducer per special case for now, as generating relevant Hypothesis arrays is tricky enough for a PR in its own right. Will explore this more later anywho.

asmeurer:
Vendoring meaning copying the test suite into the library's own test suite. It's true no one has been doing that yet, but it is something that we've supported up until now.

honno:
I resolved #40! I've documented the relevant areas, although generally there's a few small spots I'd like to cover still.

Generally I'm happy for this to get merged as-is. Here's the fairly simple TODOs I can think of for now or future PRs:

- [ ] Equivalent operator tests
- [ ] Test for non floating point special cases
- [ ] Test for functions outside the elementwise family

And to fully replace `generate_stubs.py`, I'll need to:

- [ ] Update array module stubbing
- [ ] Update signature tests

Also, I'm not quite sold it's worth supporting hypothetical situations where users might not be able to use git submodules, which _might_ be the case if the test suite was to be vendored. But in any case, I'd really like to not do this until I write a pre-commit script, lest we get out-of-sync stubs like last time.

- [ ] Commit actual `signatures` files to remove need of git submodules

array_api_tests/test_special_cases.py
```diff
@@ -469,6 +467,24 @@ def from_dtype(dtype: DataType, **kw) -> st.SearchStrategy[float]:
...
+    """
+    Parses a Sphinx-formatted result string to return:
+
+     1. A function which takes an input and returns True if it is the expected
+        result (or meets the condition of the expected result), otherwise False.
+     2. A string that expresses the result.
+
+    e.g.
+
+        >>> check_result, expr = parse_result('``42``')
```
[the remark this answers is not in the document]
asmeurer: I'm confused by the line after this one then. What is `expr_template`?

honno:
I might write a separate issue on vendoring, just to note in today's meeting we discussed this... I realise a git submodule which is already inside a folder is not certainly not ideal heh. The current options seem to be:

1. Hard code the `signatures` file ala `generate_stubs.py` 
2. Allow the user to specify somewhere (variable? env?) where the `signatures` folder/package is, so someone vendoring can vendor the spec too and use that


Generally, it seems to be a good idea to
- [ ] have the submodule on the top-level

array_api_tests/test_special_cases.py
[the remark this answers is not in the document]
honno: My bad :facepalm: I should remove the line after

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+                _x1_cond_from_dtype = shared_from_dtype  # type: ignore
+
+                def _x2_cond_from_dtype(dtype, **kw) -> st.SearchStrategy[float]:
+                    return shared_from_dtype(dtype, **kw).map(input_wrapper)
+
+            else:
+                raise ParseError(cond_str)
+
+            x1_cond_from_dtypes.append(BoundFromDtype(base_func=_x1_cond_from_dtype))
+            x2_cond_from_dtypes.append(BoundFromDtype(base_func=_x2_cond_from_dtype))
```
honno, line 857: note to self: should comment here

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+
+        def partial_cond(i1: float, i2: float) -> bool:
+            return unary_cond(input_wrapper(i1)) and unary_cond(input_wrapper(i2))
+
+    else:
+
+        def partial_cond(i1: float, i2: float) -> bool:
+            return unary_cond(input_wrapper(i1)) or unary_cond(input_wrapper(i2))
+
+    return partial_cond
```
honno, line 758: docs todo

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+    elif eq_to == BinaryCondArg.SECOND:
+
+        def check_result(i1: float, i2: float, result: float) -> bool:
+            eq = make_strict_eq(input_wrapper(i2))
+            return eq(result)
+
+    else:
+        raise ValueError(f"{eq_to=} must be FIRST or SECOND")
+
+    return check_result
```
honno, line 784: docs todo

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+
+
+def make_binary_check_result(check_just_result: UnaryCheck) -> BinaryResultCheck:
+    def check_result(i1: float, i2: float, result: float) -> bool:
+        return check_just_result(result)
+
+    return check_result
+
+
+def integers_from_dtype(dtype: DataType, **kw) -> st.SearchStrategy[float]:
```
honno, line 794: docs todo

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+        if "exclude_min" in kw.keys():
+            m += 1
+    if "max_value" in kw.keys():
+        M = kw["max_value"]
+        if "exclude_max" in kw.keys():
+            M -= 1
+    return st.integers(math.ceil(m), math.floor(M)).map(float)
+
+
+def parse_binary_case(case_str: str) -> BinaryCase:
```
honno, line 810: docs todo

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+            raise ParseError(cond_str)
+        return cond, expr_template, wrap_strat_as_from_dtype(strat)
+
+    if not_cond:
+        expr_template = f"not {expr_template}"
+        cond = make_not_cond(cond)
+        kwargs = {}
+        filter_ = cond
+    assert kwargs is not None
+    return cond, expr_template, BoundFromDtype(kwargs, filter_, from_dtype)
```
honno, line 467: comment whats happening in these "stages". also note the difference between using normal strategies and `BoundFromDtype`, and why they're mixed-and-match

array_api_tests/test_special_cases.py
```diff
@@ -0,0 +1,1209 @@
...
+            raise ParseError(case_str)
+        x1_cond_from_dtype = sum(x1_cond_from_dtypes, start=BoundFromDtype())
+    if len(x2_cond_from_dtypes) == 0:
+        x2_cond_from_dtype = xps.from_dtype
+    elif len(x2_cond_from_dtypes) == 1:
+        x2_cond_from_dtype = x2_cond_from_dtypes[0]
+    else:
+        if not all(isinstance(fd, BoundFromDtype) for fd in x2_cond_from_dtypes):
+            raise ParseError(case_str)
+        x2_cond_from_dtype = sum(x2_cond_from_dtypes, start=BoundFromDtype())
```
honno, line 1001: comment how the x1/x2 condition `from_dtype()` funcs are gathered

honno (commented):
Ok, here's the remaining areas I think I should document.

honno:
@asmeurer happy to merge on my end—not critical but would nice to get this in soon so we can get feedback from say Dask about these tests. See https://github.com/data-apis/array-api-tests/pull/104#issuecomment-1067112304 for notes on further work and vendoring.

PR #110, opened 2022-04-12
Revamped signature tests

honno:
As attempted in #109, this PR revamps the signature tests so that they try using `inspect` before passing actual arguments. This wasn't done initially due to compiled funcs not being easily inspectable https://github.com/data-apis/array-api-tests/pull/109#issuecomment-1076706621, but I think it's best if we try using them as they **give us full coverage very easily when they do work... which is most of the time!**

The fallback option of passing args looks a bit different now. All arguments are inferred, and when none can be inferred we just skip the test. It turns out generating arrays somewhat naively works in a slim majority of functions, using our `dtype_helpers.py` utils to get valid dtypes. Inferred defaults from the stub signatures work most of the time too.

In cases we can't infer arguments, we skip. It seems right now only ~2/3rds of functions can get tested via the fallback option (i.e. passing args, not using `inspect`). IMO for a solution which has no hard-coded examples, that's a pretty promising start. I had played around heuristics like custom shape and element strategies for certain functions... but it kinda got messy so I thought I'd leave it to later, as it doesn't seem high priority.

Hypothesis is not strictly necessary for the fallback option, but I like writing generalised data specifications rather than hard-coded arguments—this does end up working in our favour when it comes to specifying dtypes, and shapes/elements in the future. I set `max_examples=1`when used as we don't actually care about testing edge cases here (also worth noting data draws are much faster like this compared to using `<strat>.example()`).

`test_signatures.py` is a nasty diff heh as I basically start from scratch, you'll prob just want to look at it as-is.

PR #112, opened 2022-04-13
Implement `EqualityMapping` and use for relevant dtype helpers

### LINKED ISSUE #111: Use equality instead of hasing for dtype helpers
@asmeurer noted in https://github.com/data-apis/array-api-tests/pull/110#issuecomment-1097185020 that NumPy proper (i.e. `numpy`, as opposed to `numpy.array_api`) wasn't working with the test suite due to the [following code](
https://github.com/data-apis/array-api-tests/blob/98160117e839c4c1aa489b72a26951fbf66eccb1/array_api_tests/dtype_helpers.py#L120-L137). The problem was that the dtype attribute of NumPy-proper arrays were being used in conjuction with namespaced dtypes—these two are different objects, and thus have different hashes so they look like different keys in a `dict`.

```python
>>> import numpy as np
>>> dtypes_map = {np.int64: "foo"}
>>> dtypes_map[np.asarray(0).dtype]
KeyError
```

This behaviour is isn't specifically ruled out in the spec, as the [spec only says dtypes need equality](https://data-apis.org/array-api/latest/API_specification/data_types.html#data-type-objects), which NumPy-proper does conform to.

```python
>>> np.asarray(0).dtype == np.int64
True
```

So the test suite should phase out assumptions of namespaced dtypes and array dtypes sharing the same hash, instead relying on equality.

honno:
Resolves #111 by implementing a dict-like class `EqualityMapping` that uses equality for indexing, as opposed to hashing.

The test suite would work again with NumPy-proper, except another issue now persists if you use newer versions of Hypothesis (specifically after HypothesisWorks/hypothesis#3156)—an internal check assumes floats are being returned for `xp.finfo()`, which is not the case with NumPy-proper and so an unrelated error occurs for any test using `xps.from_dtype()`... so any test using `xps.arrays()` :sweat_smile: I wrote a quick hack which now bypasses this internal check in `__init__.py`)... it's certainly not ideal, along with my original `st.floats()` monkey patching, so will mull it over.

### DIFF
--- a/array_api_tests/dtype_helpers.py
+++ b/array_api_tests/dtype_helpers.py
@@ -1,4 +1,4 @@
-from collections import Mapping
+from collections.abc import Mapping
 from functools import lru_cache
 from typing import NamedTuple, Tuple, Union
 from warnings import warn
--- a/array_api_tests/dtype_helpers.py
+++ b/array_api_tests/dtype_helpers.py
@@ -1,6 +1,6 @@
 from collections.abc import Mapping
 from functools import lru_cache
-from typing import NamedTuple, Tuple, Union
+from typing import Any, NamedTuple, Sequence, Tuple, Union
 from warnings import warn
 
 from . import _array_module as xp
@@ -48,8 +48,8 @@ class EqualityMapping(Mapping):
     See https://data-apis.org/array-api/latest/API_specification/data_types.html#data-type-objects
     """
 
-    def __init__(self, mapping: Mapping):
-        keys = list(mapping.keys())
+    def __init__(self, key_value_pairs: Sequence[Tuple[Any, Any]]):
+        keys = [k for k, _ in key_value_pairs]
         for i, key in enumerate(keys):
             if not (key == key):  # specifically checking __eq__, not __neq__
                 raise ValueError("Key {key!r} does not have equality with itself")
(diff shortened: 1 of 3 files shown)


honno:
If you'd like to try this with the #110 changes, see https://github.com/honno/array-api-tests/tree/try-inspect-sig-eq-map (it's not a clean rebase otherwise).

array_api_tests/dtype_helpers.py
```diff
@@ -90,31 +133,37 @@ class MinMax(NamedTuple):
...
-}
-
-
-dtype_signed = {
-    **{d: True for d in int_dtypes},
-    **{d: False for d in uint_dtypes},
-}
+dtype_ranges = EqualityMapping(
+    {
+        xp.int8: MinMax(-128, +127),
```
asmeurer, line 138: These still create the dict. If a dtype object really didn't support hashing it would fail here.

honno: Ahh 🤦 I 'spose this solves our immediate problem of supporting NumPy proper. Replacing dicts here are pretty annoying, so we can leave it later as it's not high priority. 

I'll write a top-level assertion to see if all the dtype objects support hashing for now.

asmeurer: Should just be a question of using a list of tuples instead of a dictionary, no?

honno: > Should just be a question of using a list of tuples instead of a dictionary, no?

Yep, pushed a change which now uses key-value tuples for the internal mapping, and subsequently updated the respective dtype helpers.

honno:
I'll merge this tomorrow unless any blockers are identified.

PR #123, opened 2022-05-21
Prevent string dtypes from being split into chars in pytest_helpers

djl11:
For some frameworks, the data type might be represented as a string, which will then be incorrectly split into characters by the pytest helper. As an example, [Ivy](https://github.com/unifyai/ivy) uses strings to represent dtypes such that a unified framework-agnostic representation can be used.


asmeurer:
Interesting to know that there is at least one library using strings as dtypes.

Issue #125, opened 2022-05-24
Do not generate incredibly large distances for `test_linspace`


honno:
Avoids things like https://github.com/numpy/numpy/issues/21585 (currently a very rare flaky failure for our CI), which aren't really compliance issues so much as weird edge cases no one actually needs to care about.

PR #131, opened 2022-06-24
JSON reporting

asmeurer:
JSON reporting will be done with the [pytest-json-report package](https://pypi.org/project/pytest-json-report/). This produces a pretty usable JSON and makes it easy to extend it with custom metadata at test run time.

The idea will be to take the JSON generated by pytest-json-report and translate it into something more usable by a reporting tool. To start with, I will focus on test_has_names, which has been reintroduced here, to list which spec names are or are not defined in the given package.

### DIFF
--- a/array_api_tests/stubs.py
+++ b/array_api_tests/stubs.py
@@ -9,6 +9,7 @@ from typing import Dict, List
 __all__ = [
     "name_to_func",
     "array_methods",
+    "array_attributes",
     "category_to_funcs",
     "EXTENSIONS",
     "extension_to_funcs",
--- a/array_api_tests/stubs.py
+++ b/array_api_tests/stubs.py
@@ -34,6 +34,10 @@ array_methods = [
     f for n, f in inspect.getmembers(array, predicate=inspect.isfunction)
     if n != "__init__"  # probably exists for Sphinx
 ]
+array_attributes = [
+    n for n, f in inspect.getmembers(array, predicate=lambda x: not inspect.isfunction(x))
+    if n != "__init__"  # probably exists for Sphinx
+]
 
 category_to_funcs: Dict[str, List[FunctionType]] = {}
 for name, mod in name_to_mod.items():
--- a/array_api_tests/test_signatures.py
+++ b/array_api_tests/test_signatures.py
@@ -32,9 +32,9 @@ from hypothesis.strategies import DataObject
 from . import dtype_helpers as dh
 from . import hypothesis_helpers as hh
 from . import xps
-from ._array_module import _UndefinedStub
-from ._array_module import mod as xp
-from .stubs import array_methods, category_to_funcs, extension_to_funcs
+from ._array_module import _UndefinedStub, mod as xp, mod_name
+from .stubs import (array_attributes, array_methods, category_to_funcs,
+                    extension_to_funcs, EXTENSIONS)
 from .typing import Array, DataType
 
 pytestmark = pytest.mark.ci
@@ -252,3 +252,32 @@ def test_array_method_signature(stub: FunctionType, data: DataObject):
     assert hasattr(x, stub.__name__), f"{stub.__name__} not found in array object {x!r}"
     method = getattr(x, stub.__name__)
     _test_func_signature(method, stub, array=x)
+
+has_name_params = []
+for ext, stubs in extension_to_funcs.items():
+    for stub in stubs:
+        has_name_params.append(pytest.param(ext, stub.__name__))
+for cat, stubs in category_to_funcs.items():
+    for stub in stubs:
+        has_name_params.append(pytest.param(cat, stub.__name__))
+for meth in array_methods:
+    has_name_params.append(pytest.param('array_method', meth.__name__))
+for attr in array_attributes:
+    has_name_params.append(pytest.param('array_attribute', attr))
+
+# This is a very basic test to see what names are defined in a library. It
+# does not even require functioning hypothesis array_api support.
+@pytest.mark.parametrize("category, name", has_name_params)
+def test_has_names(category, name):
+    if category in EXTENSIONS:
+        ext_mod = getattr(xp, category)
+        assert hasattr(ext_mod, name), f"{mod_name} is missing the {category} extension function {name}()"
+    elif category.startswith('array_'):
+        # TODO: This would fail if ones() is missing.
+        arr = xp.ones((1, 1))
+        if category == 'array_attribute':
+            assert hasattr(arr, name), f"The {mod_name} array object is missing the attribute {name}"
+        else:
+            assert hasattr(arr, name), f"The {mod_name} array object is missing the method {name}()"
+    else:
+        assert hasattr(xp, name), f"{mod_name} is missing the {category} function {name}()"
--- a/array_api_tests/test_signatures.py
+++ b/array_api_tests/test_signatures.py
@@ -32,9 +32,8 @@ from hypothesis.strategies import DataObject
 from . import dtype_helpers as dh
 from . import hypothesis_helpers as hh
 from . import xps
-from ._array_module import _UndefinedStub, mod as xp, mod_name
-from .stubs import (array_attributes, array_methods, category_to_funcs,
-                    extension_to_funcs, EXTENSIONS)
+from ._array_module import _UndefinedStub, mod as xp
+from .stubs import array_methods, category_to_funcs, extension_to_funcs
 from .typing import Array, DataType
 
 pytestmark = pytest.mark.ci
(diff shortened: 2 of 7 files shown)


array_api_tests/stubs.py
```diff
@@ -34,6 +34,10 @@
     f for n, f in inspect.getmembers(array, predicate=inspect.isfunction)
     if n != "__init__"  # probably exists for Sphinx
 ]
+array_attributes = [
```
honno, line 37: How about `array_properties`?

honno (commented):
`pytest-json-report` looks nifty!

array_api_tests/test_signatures.py
```diff
@@ -32,8 +32,7 @@ def squeeze(x, /, axis):
 from . import dtype_helpers as dh
 from . import hypothesis_helpers as hh
 from . import xps
-from ._array_module import _UndefinedStub
-from ._array_module import mod as xp
+from ._array_module import _UndefinedStub, mod as xp
```
honno, line 35: Hmm, I'm guess you prefer this import style. I've been using `black` and `isort` locally for the newer files I introduced, so by virtue prefer how it is already :sweat_smile:

asmeurer: It does seem weird to put these on different lines, but I didn't realize that this was left over from other changes that I originally made in this file, so we can revert it.

rgommers:
@asmeurer nice! For review/discussion it would be useful if you added an example of what the JSON output looks like (maybe in a gist?) if you have one.

asmeurer:
This is ready to be merged. I've extracted a lot of custom metadata from the tests in to the report. For hypothesis errors, for now it only extracts the error message (string). I could look into trying to extract the actual data itself and JSON serializing it, but I think that will be harder, so for now I'm going to wait until it looks like we actually need that. 

I have uploaded an example report for `numpy.array_api` here https://gist.github.com/asmeurer/70d9b05e91a67d6b035b9b5305e427f0. Note that the numpy.array_api report is 11 MB uncompressed.  I also ran the tests with plain `numpy`, and the report size is 595 MB (I did not upload it). In general, many failures will balloon the file size because the full traceback is stored for each failure. However, they compress quite well (I'm not sure what the best compression to use for the web, but a basic zip compression takes the numpy.array_api report down to 375 KB and the plain numpy report down to 3.4 MB).

If there is any other metadata you can think of that should be included, let me know.

asmeurer:
Something to keep in mind when doing any further developments on the test suite is to make sure that any inputs to test functions are JSON-serializable. If they aren't, the `to_json_serializable` function in `reporting.py` should be updated. I've made it so that dataclasses and namedtuples are serialized as objects (dicts), and functions are serialized to a string of their name, but anything that is unrecognized will just be serialized as its `repr`.

rgommers:
> In general, many failures will balloon the file size because the full traceback is stored for each failure.

Maybe using `--tb=native` is helpful? Those long tracebacks are usually unhelpful, and pytest has a terrible default there.

asmeurer:
OK, it turns out I was actually wrong here. The tracebacks do contribute a some amount of data in general, but it's not actually the reason the plain numpy report is over half a gigabyte. That's apparently because of the warnings. There are over a million warnings emitted in the run, and they account for 98% of the JSON file size. Of these 98% are the warning "`np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here." (I guess it is getting emitted for every single hypothesis run).

Here's the full warning metadata that's repeated so many times:

```json
{
    "message": "`np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.\nDeprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations",
    "category": "DeprecationWarning",
    "when": "runtest",
    "filename": "/Users/aaronmeurer/anaconda3/envs/array-apis/lib/python3.9/site-packages/hypothesis/extra/array_api.py",
    "lineno": 119
}
```

@honno, any idea why this warning would be omitted every single time hypothesis calls `np.bool`, instead of just once?

I did try switching from the default traceback type to native, but the savings are actually pretty small, so I'm going to revert it because the default type has more information, and because it keeps the code simpler.

honno:
> @honno, any idea why this warning would be omitted every single time hypothesis calls `np.bool`, instead of just once?

(I assume you're testing the top-level `numpy` namespace)

If you meant _emitted_, I'd guess it's to do with the pytest reporting steps. Hypothesis might "merge" warnings together as it does errors in the reporting stage, so pytest would only show one warning, but however we capturing warnings here ignores that. Generally it makes sense that warning is emitted many times, as `np.bool` is accessed many times due to the vast number of Hypothesis examples generated in a run of the whole test suite.

honno:
Note flaky CI failure is due to https://github.com/data-apis/array-api-tests/issues/125

conftest.py
```diff
@@ -7,8 +7,9 @@
 from array_api_tests import _array_module as xp
 from array_api_tests._array_module import _UndefinedStub
 
-settings.register_profile("xp_default", deadline=800)
+from reporting import pytest_metadata, add_extra_json_metadata # noqa
```
honno, line 10: Could you add a small comment explaining what's being imported here and why? I assume it's doing some magic when imported.

asmeurer: Yes, the names have to be defined here so that pytest will see them as fixtures. I decided to keep the reporting in a separate file to keep things clean, since it's already a good bit of code and could potentially be more if we decide to add more stuff to the report in the future.

asmeurer:
> personally I'd prefer test_has_names.py:: test_has_names -> test_has_attributes.py:: test_has_attributes

Your suggestion is to have two separate files for this? I don't really see the value in that, especially since the file is so short and uses the same logic for names and attributes.

asmeurer:
My point with the warnings is that by default, warnings are only omitted once per warning:

```
>>> import numpy as np
>>> np.bool
<stdin>:1: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
<class 'bool'>
>>> np.bool
<class 'bool'>
```

But this exact same warning is emitted every single time. So somewhere something is changing the warnings filter to enable them always. I did see some `warnings` code in hypothesis but I wasn't sure if it was what was causing this. I guess it could also be pytest doing it somehow.

honno:
> > personally I'd prefer test_has_names.py:: test_has_names -> test_has_attributes.py:: test_has_attributes
> 
> Your suggestion is to have two separate files for this? I don't really see the value in that, especially since the file is so short and uses the same logic for names and attributes.

Ah I was being lazy heh, I meant rename `test_has_names.py` to `test_has_attributes.py`, and rename the test case `test_has_names` to `test_has_attributes`.

honno:
> 

Ooo, this feels like Numpy is doing some magic which pytest/Hypothesis isn't equipped for?

rgommers:
> Ooo, this feels like Numpy is doing some magic which pytest/Hypothesis isn't equipped for? Probably requires some hacking around to fix.

It shouldn't do anything fancy; the one thing is `filterwarnings = error` in `pytest.ini`.

asmeurer:
NumPy's pytest.ini is irrelevant here. It looks like the pytest defaults are to store each warning invocation separately. Since the warning metadata doesn't even contain enough information to distinguish the different warnings, I think the simplest solution here is to post-process the JSON to deduplicate the warnings. (probably a lot of the stuff I'm doing here could be upstreamed to the pytest-json-report library)

asmeurer:
OK, I deduplicated the warnings output. The plain numpy report is now 11 MB and numpy.array_api is 6.1 MB (both uncompressed). I've uploaded examples of each to the [gist](https://gist.github.com/asmeurer/70d9b05e91a67d6b035b9b5305e427f0). I'd like to hear feedback from Athan on whether I need to try to reduce this file size any further. Again, these are uncompressed sizes, and they compress well (both are under 0.5 MB after basic zip compression).

honno:
> > Ooo, this feels like Numpy is doing some magic which pytest/Hypothesis isn't equipped for? Probably requires some hacking around to fix.
> 
> It shouldn't do anything fancy; the one thing is `filterwarnings = error` in `pytest.ini`.

Ah my confusion here is say the following

```python
# Using ipython
>>> class foopy_factory:
...     @property
...     def bool(self):
...         warn("don't use this", DeprecationWarning)
...         return bool
... 
>>> foopy = foopy_factory()
>>> foopy.bool
DeprecationWarning: don't use this
bool
>>> foopy.bool
DeprecationWarning: don't use this  # warning is emitted again
bool
```

the warning is emitted for each access of `foopy.bool`, but Aaron's example only show it emit the first time.

> ```python
> >>> import numpy as np
> >>> np.bool
> <stdin>:1: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
> Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
> <class 'bool'>
> >>> np.bool
> <class 'bool'>
> ```

I see now that the emit-once behaviour is seen using the default `python` interpreter, but not using say IPython.

```python
# Using python
>>> class foopy_factory:
...     @property
...     def bool(self):
...         warn("don't use this", DeprecationWarning)
...         return bool
... 
>>> foopy = foopy_factory()
>>> foopy.bool
DeprecationWarning: don't use this
bool
>>> foopy.bool
bool  # no warning this time
```

Seems like Aaron's found a solution anywho, but might be worth noting that environment things can affect this behaviour... or I might be missing something, I'm pretty ignorant on this stuff heh.

asmeurer:
Need to have this merged so I can start testing it.

kgryte:
@asmeurer Looking through the sample JSON, they look fine to me. No need to reduce further, IMO. The raw files can be parsed and transformed for subsequent presentation on, e.g., a website. So not too concerned regarding initial report file sizes, particularly as we can store them on disk in compressed format.

kgryte:
One thing we should consider, however, is versioning the JSON format. E.g., using [JSON schema](http://json-schema.org/). This would allow versioning our backend server APIs to handle transformations of different versions of the JSON report.

PR #136, opened 2022-08-09
MAINT: mapping f-strings

tylerjereddy:
* `EqualityMapping` was missing some f-strings that I needed
while testing the conformance of the [`pykokkos`](https://github.com/kokkos/pykokkos/tree/develop) library with
the array API standard


honno (approved):
LGTM, thanks!

On our end, it might be worth adding a (very conservatively configured) style linter to our `pre-commit` pick up on this kind of stuff.

PR #141, opened 2022-09-29
Filter out large distances in `test_linspace`

### LINKED ISSUE #125: Do not generate incredibly large distances for `test_linspace`

honno:
Resolves #125

### DIFF
--- a/array_api_tests/test_creation_functions.py
+++ b/array_api_tests/test_creation_functions.py
@@ -445,6 +445,9 @@ def test_linspace(num, dtype, endpoint, data):
     # avoid overflow errors
     assume(not xp.isnan(xp.asarray(stop - start, dtype=_dtype)))
     assume(not xp.isnan(xp.asarray(start - stop, dtype=_dtype)))
+    # avoid generating very large distances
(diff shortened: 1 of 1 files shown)


asmeurer:
I suppose this is different from https://github.com/numpy/numpy/issues/18881. 

This looks good. One could probably get more fine grained by allowing large start and stop as long as they are themselves close to one another (look at the ratio start/stop), but it's probably not a huge deal either way.

honno:
> I suppose this is different from [numpy/numpy#18881](https://github.com/numpy/numpy/issues/18881).

Ayup, this relates to https://github.com/numpy/numpy/issues/21585


> This looks good. One could probably get more fine grained by allowing large start and stop as long as they are themselves close to one another (look at the ratio start/stop), but it's probably not a huge deal either way.

I think this does still generate such cases. Only the distance between start and stop is checked (after-the-fact), rather than having limits on start and stop.

honno:
Seems uncontroversial and is nice for our CI, so merging.

Issue #20, opened 2021-09-20
Split the test suite for spec versions


asmeurer:
The 2021 version of the spec is going to be finalized soon. The test suite isn't finished yet. However, we want to have a version of the test suite that runs only against 2021, so that people can run the suite without having to worry about features that are in the 2022 draft.

We should figure out some way to enable or disable test suite features. It would also be nice to be able to do this for the extensions, so that people can easily run the test suite without the linear algebra extension, for instance. Perhaps the best way to do this is using pytest marks? Another idea would be to split the parts of the suite into submodules so that people can run `pytest array_api_tests/2021` or `pytest array_api_tests/2021/extensions/linear_algebra`. For vendoring people could remove any submodules that aren't relevant. 

I'm guessing pytest marks will be better, but I need to investigate them to see how well suited they are for this.

honno:
Something useful I learnt recently—we can specify markers for all test methods in a module via assigning the `pytestmark` variable once per-file.
```python
pytestmark = [...]
```

honno:
I'll formalise my thoughts later, just to say we added versioning support to Hypothesis in https://github.com/HypothesisWorks/hypothesis/pull/3456, which I'll probably take inspiration from when working on this issue.

PR #144, opened 2022-09-30
Bump Hypothesis, specify `api_version` in `make_strategies_namespace()`

honno:
This PR prevents Hypothesis from inferring versions by just specifying `2021.12`, given the test suite still only assumes `2021.12` features. This is a temporary solution to work with a recent Hypothesis update, hopefully to prevent issues like https://github.com/cupy/cupy/pull/7072. Next step is to support testing different versions (#20), where instead of hardcoding this, we can allow specification via an env variable.

### DIFF
--- a/array_api_tests/__init__.py
+++ b/array_api_tests/__init__.py
@@ -41,7 +41,7 @@ except AttributeError:
     pass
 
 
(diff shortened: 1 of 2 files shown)


honno:
Merging because it's pretty simple and critical for CI (ours and others).

PR #146, opened 2022-10-01
database functionality(optional) for storing hypothesis cache

Aarsh2001:
Hi, these changes enable the Array API test suite to be run as part of a CI solution for an adhering framework, with full hypothesis features including database usage, which are essential for maximum the effectiveness and efficiency of the testing across multiple commits.


honno (commented):
Thanks for the PR, but unfortunately I don't see how this will get in.

First off, do you really need a database? CI consistency is nice but really more often than not Hypothesis will raise the same errors consistently, and then any flaky test you should be skipping anyway so it doesn't affect local development. The maintence effort here doesn't seem worth.

In any case, I don't want to add some super niche thing to this repo, given it should really live beyond Aaron and I, and changes like this are pretty mystifying... heck I maintain Hypothesis and I have no idea how database things work!

Also as an aside, make sure you don't commit editor-specific things into PRs (i.e. `idea/workspace.xml` here), unless that's the workflow of the repo.

asmeurer:
I would prefer if this sort of thing could be done without having to edit conftest.py. Maybe that's a feature request for hypothesis. Anyway, if we do add this, I don't think it should be added in requirements.txt.

Aarsh2001:
Can we not do pytest command line arguments for this? In that case i don't think there would be any hardcoding, looking forward to your thoughts?

honno:
> Can we not do pytest command line arguments for this?, have another `CI flag` maybe. In that case i don't think there would be any hardcoding, although it would still change `conftest`. Looking forward to your thoughts.

IMO it's not a question of if we can, but if we should. [As I said](https://github.com/data-apis/array-api-tests/pull/146#pullrequestreview-1127608139), support for utilising Redis databases is an inherently mystifying feature, where this repo is already mystifying enough heh. I really don't want to introduce things I don't understand, nor will probably ever care to understand.

asmeurer:
Surely there must be a way to do this without modifying a project's specific conftest.py file. If there isn't, I would make an upstream request to the hypothesis project about it.

PR #149, opened 2022-10-06
Fix array_attributes

asmeurer:
Previously it was including every method inherited from object, like __dict__. Now it only includes the properties actually defined on the array stub class.


honno (approved):
Oh yeah, much nicer.

PR #155, opened 2022-11-23
Remove type promotion tests

### LINKED ISSUE #143: Remove `test_type_promotion.py`
To summarise previous discussions as far as I can tell:
* We want to test type promotion anyway in our primary tests.
* There's still value of type promotion for developers of compliant libraries, given it pinpoints the issues.

The fact is though, I don't care about maintaining it. We already have a hacky [`conftest.py`](https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/conftest.py) just to skip some flaky cases we identified, and I don't see it worth my time supporting complex cases. Testing type promotion isn't simple for a lot of functions, given other rules for inputs (namely broadcasting).

I think leaning on the primary tests solely is just a more sane approach anyway, as whilst bugs can sometimes be lost in the noise, they're usually getting reported and easier to identify once you fix other bugs you'll have to fix anyway.

honno:
Resolves #143. I'm exploring complex support and it feels even more weird to have these tests, given the machinery we'd need for #20.


kgryte:
What would be the strategy for testing type promotion compliance across spec versions moving forward?

honno:
> What would be the strategy for testing type promotion compliance across spec versions moving forward?

Our primary tests cover type promotion! I have little motivation maintaining tests specific to type promotion given they already get covered and any function/method testing is pretty finnicky to setup.

asmeurer:
It's still likely that at some point in the future we'll want some way to distinguish different parts of the test suite. That extends beyond just type promotion, though. 

I agree that test_type_promotion is redundant at this point, so we should be OK with removing it.

PR #156, opened 2022-11-30
Don't require can_cast to return False for cross-kind casting

### LINKED ISSUE #151: test_can_cast should not disallow cross-kind casting
`test_can_cast` currently fails with NumPy because NumPy allows cross-kind casting. However, the spec only says that mixed-kind casting is undefined (https://data-apis.org/array-api/latest/API_specification/type_promotion.html), and `can_cast` just says it should follow the type promotion rules (https://data-apis.org/array-api/latest/API_specification/generated/signatures.data_type_functions.can_cast.html). So it should not test this case. If anything, it should check that if can_cast does return True for a mixed-kind cast, that such a cast is actually allowed on an array of that dtype.

asmeurer:
Fixes #151.

### DIFF
--- a/array_api_tests/test_data_type_functions.py
+++ b/array_api_tests/test_data_type_functions.py
@@ -124,7 +124,11 @@ def test_can_cast(_from, to, data):
             expected = from_min >= to_min and from_max <= to_max
         else:
             expected = False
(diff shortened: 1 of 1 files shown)


honno (approved):
LGTM!

Issue #90, opened 2022-02-02
Skip rather than xfail tests in workflows


honno:
Currently our NumPy workflow has a flaky test in `test_creation_functions.py::test_linspace`, which means sometimes we'll get an `XFAIL` and sometimes an `XPASS`. The random-ish nature of Hypothesis means this is often a possibility for bugs the test suite identifies. Therefore I think instead of xfailing tests for a workflow, we should skip them completely. We could also mix-and-match, but I think it's best to set a precedent that's simple for outsiders to understand and use.

asmeurer:
Maybe implementers like @tomwhite could comment here. My impression is that XFAILs are more useful than outright skips as they can alert you that the test should be removed from the file when the feature in question is implemented/fixed. 

I agree that false positives are a problem, so you might not necessarily want to use the pytest flag that fails the tests if there is an xpass. I also wonder if it's possible to automatically increase the number of max iterations for XFAIL so that if it does xpass, it will do so in a less flaky way.

asmeurer:
At best we can have both skips.txt and xfails.txt. Personally I think we should be xfailing rather than skipping the NumPy tests we run on our CI, for the reason I mentioned above.

honno:
Yeah the utility of xfails are something I wouldn't want to lose, but this seems a hard limitation with Hypothesis. I had actually written this issue after I saw Hypothesis CI having flaky tests skipped instead of xfailed.

> At best we can have both skips.txt and xfails.txt..

I think the burden of distinguishing between what is a flaky test and what test will fail is too great for folks not well-versed in how Hypothesis works, like some of the array library developers I've seen use the suite. Split skips and xfails does make CI adoption more confusing/cumbersome too.

> I also wonder if it's possible to automatically increase the number of max iterations for XFAIL so that if it does xpass, it will do so in a less flaky way.

I imagine it would be way too much effort for something that is still not that stable.

asmeurer:
But the flakyness is only a concern if you actually configure your tests to fail when there are XPASSes. Otherwise, it just shows the XPASS fails in the log. They might differ occasionally from run to run, but that's not a huge deal. But the point is that you can notice this if you pay attention to the build, or if you run the tests locally. With skips, there's no way to notice it at all. 

[...]

PR #157, opened 2023-01-03
Add support for an xfails file, and allow the files to be specified with a flag

asmeurer:
I think the existing file for NumPy on CI needs to be updated, but I haven't looked into this. We really ought to prefer XFAIL to skips everywhere so we know when a test starts to pass. Note that XPASS doesn't actually cause a failure by default unless `-o xfail_strict=True` is used.


array_api_tests/test_has_names.py
```diff
@@ -9,6 +9,8 @@
 from .stubs import (array_attributes, array_methods, category_to_funcs,
                     extension_to_funcs, EXTENSIONS)
 
+pytestmark = pytest.mark.ci
```
honno, line 12: These tests feel a bit redundant for downstream CI runs, given the primary tests cover this implicitly, so to me it just adds noise to get the same failures twice.

honno:
I still disagree we should support xfails. Please address all the points made in https://github.com/data-apis/array-api-tests/issues/90#issuecomment-1054271996

array_api_tests/test_has_names.py
asmeurer: If one of these tests fails that means the function or attribute doesn't exist at all. The error message here will be quite clear on that, whereas the error message in the other tests might not be as clear, especially for missing attributes. 

Also, it's relevant because I want to include these failures as part of the xfails so that I know when they are fixed.

asmeurer:
I need this for the array api compat library. I want to be alerted on CI when an xfailing test starts passing, so that I know that some behavior is working again. I'd also ideally like to do the same for the base libraries so that I know when some piece of compat code can be removed.

I'm really not worried about flakyness. If it becomes a problem, we can figure out how to fix it when we get to it. Flakyness is an inherent problem when using hypothesis, but I don't think it's a good excuse to throw good testing practices out the window. There are several ways it can be addressed, e.g., increasing max examples, storing the examples database persistently on CI, modifying the test itself in some way, or even just making that single test a skip (as opposed to only using skips for all tests).

I'm still concerned about our use of skips in this repo. Right now the NumPy CI on this repo is failing anyway, but ideally, I'd like to run both numpy and numpy.array_api, either on this repo or elsewhere, with xfails. Firstly, that will alert us to changes in the library support, which is important for numpy since we are effectively in charge of that support. And secondly, the CI on this repo currently only runs against numpy.array_api with skips. That means that those skipped tests are *never* run on CI. If they were XFAIL they would at least run at all (XFAIL with expected failure message might be more robust here, but OTOH *that* might actually be too flaky to deal with).

asmeurer:
Also I should note that I want to do the same for cupy, which I can only run locally occasionally on my PC (unless we can figure out a CI solution there, but I really doubt that will happen). Basically whenever I'm thinking about it I can pull it up and run them, or someone else with access to a CUDA GPU can. I really don't care about flakyness for that. I can afford to just run with more examples or just run the test again, given how rarely I would do it.

asmeurer:
Also, as I noted, XPASS only fails if you tell it to with `-o xpass_strict=true`. Otherwise, xfail works just like skip, except the test is actually run so you have a chance to see if it passed or not if you look at the test summary.

honno:
So if you want to use this test suite as a primary test suite to rely upon, I see the utility of xfails. Can we not promote this in the README though? I absolutely do not want to deal with flakiness issues downstream users will have down the road heh.

README.md
```diff
@@ -200,39 +209,81 @@ array_api_tests/test_add[__iadd__(x, s)]
...
+        path: your-array-library
+
+    - name: Checkout array-api-tests
+      uses: actions/checkout@v3
+      with:
+        repository: data-apis/array-api-tests
+        submodules: 'true'
+        path: array-api-tests
+
+    - name: Run the array API test suite
```
honno, line 245: Do we need to be verbose here? I'd rather this example be minified to what we actually want to demonstrate, keeping the README more skimmable, and not confusing alternative approaches to running this test suite on CI.

.github/workflows/numpy.yml
```diff
@@ -28,51 +28,4 @@ jobs:
...
-        array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(x1_i is +infinity and isfinite(x2_i) and x2_i < 0) -> -infinity]
-        array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(x1_i is -infinity and isfinite(x2_i) and x2_i > 0) -> -infinity]
-        array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(x1_i is -infinity and isfinite(x2_i) and x2_i < 0) -> +infinity]
-        array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(isfinite(x1_i) and x1_i > 0 and x2_i is -infinity) -> -0]
-        array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(isfinite(x1_i) and x1_i < 0 and x2_i is +infinity) -> -0]
-
-        EOF
-
-        pytest -v -rxXfE --ci
+        pytest -v -rxXfE --ci --skips-file numpy-skips.txt
```
honno, line 31:
```suggestion
        pytest -v -rxXfE --ci --skips-file numpy-skips.txt
```

I kinda liked the `cat`-in-the-yaml approach :sweat_smile: No biggie, and also I get its a bit hacky, I just like keeping the skips (+ xfails) in one file.

README.md
```diff
@@ -200,39 +209,81 @@ array_api_tests/test_add[__iadd__(x, s)]
...
 [Hypothesis](https://hypothesis.readthedocs.io/en/latest/). You can configure
-how many examples are generated using the `--max-examples` flag, which defaults
-to 100. Lower values can be useful for quick checks, and larger values should
-result in more rigorous runs. For example, `--max-examples 10_000` may find bugs
-where default runs don't but will take much longer to run.
+how many examples are generated using the `--max-examples` flag, which
+defaults to `100`. Lower values can be useful for quick checks, and larger
+values should result in more rigorous runs. For example, `--max-examples
+10_000` may find bugs where default runs don't but will take much longer to
+run.
```
honno, line 286: Seems unnecessary noise on the diff, esp. with lines already 80-or-less chars. Or is the cadence 79-or-less? in which case I see the second-to-last line was 80 chars long.

conftest.py
```diff
@@ -47,7 +48,17 @@ def pytest_addoption(parser):
...
+    )
+    parser.addoption(
+        "--skips-file",
+        action="store",
+        help="file with tests to skip. Defaults to skips.txt"
+    )
+    parser.addoption(
+        "--xfails-file",
+        action="store",
+        help="file with tests to skip. Defaults to xfails.txt"
```
honno, line 61: I like defaulting the files, but wondering if we can default it here rather than below, e.g. specify a default like `pathlib.Path(__file__).parent / "skips.txt"`... kinda rusty on argparse things myself heh.

README.md
asmeurer: There is a change here (quotes around the `100`), but GitHub's word-diff is kind of crappy. I generally try to only M-q paragraphs that I actually edit.

asmeurer:
So I've toned back the xfail stuff. You're right that there is a lot of flakyness (especially with pytorch which only seems to fail a lot of special cases tests for very specific values; probably due the way it does scalar type promotion). I do still think that actually running the test is better than skipping, but xpass isn't very usable.

In my local testing with cupy, I've come to the conclusion that xpass can work if you sufficiently populate the examples database (e.g., by running the tests with 10000 max examples). I'm hoping to figure out how to persist the examples database on CI. I know that the examples database does get invalidated sometimes when hypothesis or the test suite updates, but I expect that level of flakyness will be much less annoying.

.github/workflows/numpy.yml
asmeurer: If you look at array-api-compat, the xfails and skips files are quite large, with comments on each xfail for why it is there. It's much more maintainable to keep them in a separate file.

README.md
asmeurer: What are the alternative approaches? 

Personally as a user I would find it more helpful to have something I can just copy-paste that works. 

Also IMO this is the simple version :) Compare this to what I'm actually using in array-api-compat with a reusable workflow https://github.com/data-apis/array-api-compat/blob/main/.github/workflows/array-api-tests.yml

.github/workflows/numpy.yml
asmeurer: Oh I see this is a comment for the CI file in this repo, not the README. I still think a separate file is better. I don't like giant heredocs in CI config files.

asmeurer:
Assuming the test failures go away after merging master, any objections to merging this?

honno:
Pushed a commit to fix wildcard ids not matching the respective "child" ids, e.g.

```console
$ cat skips.txt
array_api_tests/test_operators_and_elementwise_functions.py::test_abs
$ pytest array_api_tests/test_operators_and_elementwise_functions.py::test_abs -v
array_api_tests/test_operators_and_elementwise_functions.py::test_abs[abs] SKIPPED (skips file (/home/honno/gdrive/GitHub/array-...) [ 50%]
array_api_tests/test_operators_and_elementwise_functions.py::test_abs[__abs__] PASSED
```

README.md
```diff
@@ -207,39 +216,81 @@ array_api_tests/test_add[__iadd__(x, s)]
...
+Here is an example GitHub Actions workflow file, where the xfails are stored
+in `array-api-tests.xfails.txt` in the base of the `your-array-library` repo.
+
+If you want, you can use `-o xfail_strict=True`, which causes XPASS tests (XFAIL
+tests that actually pass) to fail the test suite. However, be aware that
+XFAILures can be flaky (see below, so this may not be a good idea unless you
+use some other mitigation of such flakyness).
+
+If you don't want this behavior, you can remove it, or use `--skips-file`
+instead of `--xfails-file`.
```
honno, line 228: Much prefer highlighting `skips.txt`, and then talk about `xfails.txt`.

honno (commented):
Think this is nearly there, mostly my personal preference now.

README.md
```diff
@@ -207,39 +216,81 @@ array_api_tests/test_add[__iadd__(x, s)]
...
-        array_api_tests/test_add[__iadd__(x1, x2)]
-        array_api_tests/test_add[__iadd__(x, s)]
-        array_api_tests/test_set_functions.py
-
-        EOF
-
-        pytest -v -rxXfE --ci
+        export PYTHONPATH="${GITHUB_WORKSPACE}/array-api-compat"
+        cd ${GITHUB_WORKSPACE}/array-api-tests
+        pytest -v -rxXfE --ci --xfails-file ${GITHUB_WORKSPACE}/your-array-library/array-api-tests-xfails.txt array_api_tests/
```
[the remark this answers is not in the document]
asmeurer: The "compat" thing is definitely a typo, but in general, this is a recipe for storing the skips in the array library repo and testing against the test suite. The point of GITHUB_WORKSPACE is to show how to do that (most people don't clone multiple repos in a single GitHub action, so it's not obvious how to do it). Again, I'd prefer to have a mostly working example so that people can just copy-paste it, rather than omitting stuff that they'll just have to figure out themselves. I always prefer it when people do this in other contexts which is why I've done it here.

README.md
asmeurer: We're just not going to get agreement on this. XFAILs are better than skips. Period. It's better to run the test so you can still see if it actually failed or not. The only exception is cases where the test hangs or crashes the testsuite.

asmeurer:
Need this for testing the compat library against 2022.12, so going to merge.

PR #164, opened 2023-01-16
Bump Hypothesis to `6.21.1`

### LINKED ISSUE #159: pytorch fails signature tests with odd error message
This was discussed on slack. There are errors like 

```
FAILED array_api_tests/test_signatures.py::test_func_signature[max] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.uint8 with fill value 0
FAILED array_api_tests/test_signatures.py::test_func_signature[mean] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.float32 with fill value 0.0
FAILED array_api_tests/test_signatures.py::test_func_signature[min] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.uint8 with fill value 0
FAILED array_api_tests/test_signatures.py::test_func_signature[prod] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.uint8 with fill value 0
FAILED array_api_tests/test_signatures.py::test_func_signature[std] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.float32 with fill value 0.0
FAILED array_api_tests/test_signatures.py::test_func_signature[sum] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.uint8 with fill value 0
FAILED array_api_tests/test_signatures.py::test_func_signature[var] - hypothesis.errors.InvalidArgument: Could not create full array of dtype=torch.float32 with fill value 0.0
```

which are not easy to read 

The problem is that hypothesis tries to pass an integer to `full`, but torch only supports a tuple. I think a simple fix would be if hypothesis only tried passing a tuple to `full` to avoid this issue.

honno:
Resolves #159, as Hypothesis now doesn't pass int-shapes to `xp.full()`, which doesn't work with PyTorch.

Some other issues with the signature tests now, will explore.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pytest
 pytest-json-report
-hypothesis>=6.55.0
+hypothesis>=6.62.1
 ndindex>=1.6


honno:
Merging as trivial change that's nice to get in.

PR #174, opened 2023-03-15
`test_isdtype`

honno:
New territory here, but is currently passing for https://github.com/data-apis/array-api-compat/pull/32

### DIFF
--- a/array_api_tests/test_data_type_functions.py
+++ b/array_api_tests/test_data_type_functions.py
@@ -177,6 +177,33 @@ def test_iinfo(dtype):
     # TODO: test values
 
 
+def atomic_kinds() -> st.SearchStrategy[Union[DataType, str]]:
+    return xps.scalar_dtypes() | st.sampled_from(list(dh.kind_to_dtypes.keys()))
+
+
+@pytest.mark.min_version("2022.12")
+@given(
+    dtype=xps.scalar_dtypes(),
+    kind=atomic_kinds() | st.lists(atomic_kinds(), min_size=1).map(tuple),
+)
+def test_isdtype(dtype, kind):
+    out = xp.isdtype(dtype, kind)
+
+    assert isinstance(out, bool), f"{type(out)=}, but should be bool [isdtype()]"
+    _kinds = kind if isinstance(kind, tuple) else (kind,)
+    expected = False
+    for _kind in _kinds:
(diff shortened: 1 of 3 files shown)


array_api_tests/test_data_type_functions.py
```diff
@@ -174,6 +175,33 @@ def test_iinfo(dtype):
...
+        assert isinstance(out, bool), f"{type(out)=}, but should be bool [isdtype()]"
+        _kinds = kind if isinstance(kind, tuple) else (kind,)
+        expected = False
+        for _kind in _kinds:
+            if isinstance(_kind, str):
+                if dtype in dh.kind_to_dtypes[_kind]:
+                    expected = True
+                    break
+            else:
+                if dtype == _kind:
```
asmeurer, line 199: This implicitly assumes dtype equality works, meaning if it's broken or incorrect this might not notice. You can see the test I added in my compat PR compare the known value using dtype string names. Maybe it isn't so important here since we also test `dtype.__eq__` anyway.

array_api_tests/test_data_type_functions.py
```diff
@@ -174,6 +175,33 @@ def test_iinfo(dtype):
     # TODO: test values
 
 
+if api_version >= "2022.12":
```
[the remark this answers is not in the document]
honno: Yeah, doesn't feel good. I (seemingly) have to do this for a test like `test_conj` because we're using the `>=2022.12` feature of Hypothesis `xps.complex_dtypes()`,

https://github.com/data-apis/array-api-tests/blob/a533680532bf79cfd6aebbe65a202dc458e7ecfd/array_api_tests/test_operators_and_elementwise_functions.py#L927-L934

so kinda just did this to be consistent with the pattern. But I think the ultimate goal with `test_conj`/etc. would be to not use this pattern, so now skipping this test instead in my force-push.

array_api_tests/test_data_type_functions.py
honno: Yeah. I think we're just out of luck if that's the case. Fortunately this would get checked via a sanity check I wrote in `EqualityMapping`

https://github.com/data-apis/array-api-tests/blob/a533680532bf79cfd6aebbe65a202dc458e7ecfd/array_api_tests/dtype_helpers.py#L62-L63

PR #176, opened 2023-03-24
Use keyword-only arguments for the pytest helpers functions

asmeurer:
This makes the inscrutable helper function calls more readable, and fixes at least one instance where the arguments were called in the wrong order.

This also changes **kw to kw={} for the function keyword arguments, so there is no ambiguity between function keyword arguments and arguments to the helper function.

I've checked the codebase with pylint to make sure I didn't miss anything here, but if you know of any other tools that can check this we can run those as well.

### DIFF
--- a/array_api_tests/meta/test_pytest_helpers.py
+++ b/array_api_tests/meta/test_pytest_helpers.py
@@ -5,18 +5,18 @@ from .. import pytest_helpers as ph
 
 
 def test_assert_dtype():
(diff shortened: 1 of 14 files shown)


honno (approved):
Thanks makes sense, I really should of done this in the first place—non-defaulted kw-only args have been an awesome revelation for me!

(Generally I'm not a fan of lines exceeding ~100 columns, but if we're not going to introducing a formatter than it's just bikeshedding to address.)

PR #177, opened 2023-04-11
Specify/infer arguments for testing uninspectable signatures

### LINKED ISSUE #175: Signature tests not catching missing keyword arguments
For array_api_compat.torch, the `triu` function is missing the `k` keyword argument. But none of the tests catch this:

```
array_api_tests/test_has_names.py::test_has_names[creation-triu] PASSED                                                                                           [ 50%]
array_api_tests/test_signatures.py::test_func_signature[triu] SKIPPED (Signature for triu() is not inspectable and is too troublesome to test for otherwise)      [100%]
```

This is a great example of something that the old signature tests would handle just fine. I just tested it in 2022.03.22, the version before #110 was merged, and it finds the error just fine. 

```
E               AssertionError: Unexpected exception TypeError("triu() got an unexpected keyword argument 'k'"): triu() should accept the keyword-only argument 'k'
```

So let's please just reintroduce these old signature tests as a fallback. They are much better than nothing and hardly "too troublesome". We are missing real bugs because of this.

honno:
Should resolve #175 by re-introducing manual specification of arguments when testing uninspectable signatures.

Differences with pre-#110 `test_signatures.py`:

* Declarative approach to specifying manual arguments.
  * Arguments are explicitly specified for the functions/methods that use them, preventing assumptions leaking.
* Leverages [`inspect.Parameter.default`](https://docs.python.org/3/library/inspect.html#inspect.Parameter.default) to infer the majority of manual arguments.
* Tests all valid combinations of passing pos-or-kw arguments

### DIFF
--- a/array_api_tests/dtype_helpers.py
+++ b/array_api_tests/dtype_helpers.py
@@ -1,8 +1,8 @@
 import re
+from collections import defaultdict
 from collections.abc import Mapping
 from functools import lru_cache
-from inspect import signature
-from typing import Any, Dict, NamedTuple, Sequence, Tuple, Union
+from typing import Any, DefaultDict, NamedTuple, Sequence, Tuple, Union
 from warnings import warn
 
 from . import _array_module as xp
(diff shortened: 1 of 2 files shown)


honno:
This is nice to get in so merging.

PR #181, opened 2023-04-19
Fix some issues

asmeurer:
Some issues that came up with testing pytorch against the 2022 spec (this doesn't address any of the issues I opened).

### DIFF
--- a/array_api_tests/test_array_object.py
+++ b/array_api_tests/test_array_object.py
@@ -250,8 +250,8 @@ def make_param(method_name: str, dtype: DataType, stype: ScalarType) -> Param:
 @pytest.mark.parametrize(
     "method_name, dtype, stype",
     [make_param("__bool__", xp.bool, bool)]
-    + [make_param("__int__", d, int) for d in dh.all_int_dtypes]
-    + [make_param("__index__", d, int) for d in dh.all_int_dtypes]
+    + [make_param("__int__", d, int) for d in dh._filter_stubs(*dh.all_int_dtypes)]
+    + [make_param("__index__", d, int) for d in dh._filter_stubs(*dh.all_int_dtypes)]
     + [make_param("__float__", d, float) for d in dh.float_dtypes],
 )
 @given(data=st.data())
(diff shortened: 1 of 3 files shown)


array_api_tests/test_array_object.py
```diff
@@ -250,8 +250,8 @@ def make_param(method_name: str, dtype: DataType, stype: ScalarType) -> Param:
 @pytest.mark.parametrize(
     "method_name, dtype, stype",
     [make_param("__bool__", xp.bool, bool)]
-    + [make_param("__int__", d, int) for d in dh.all_int_dtypes]
-    + [make_param("__index__", d, int) for d in dh.all_int_dtypes]
+    + [make_param("__int__", d, int) for d in dh._filter_stubs(*dh.all_int_dtypes)]
+    + [make_param("__index__", d, int) for d in dh._filter_stubs(*dh.all_int_dtypes)]
```
asmeurer, line 254: Let me know if there's a better way to do this. I'm not sure if this internal function is supposed to be used outside of the dtype_helpers module. I'm a little unclear why `all_int_dtype` isn't just filtered to begin with?

honno: Alternative approach which skips undefined stubs

```diff
def test_scalar_casting(method_name, dtype, stype, data):
+    if isinstance(dtype, xp._UndefinedStub):
+        pytest.skip(f"{dtype} not in xp namespace")
```

I could go either way with this. I like the idea of having the test case for `uint{16,32,64}` just exist as it is expected behaviour in the spec afterall, but I also get its noisy.

> I'm a little unclear why `all_int_dtype` isn't just filtered to begin with?

IIRC it's for the above reason, so it exists for such test cases and we can subsequently skip. Again, could go either way on that now.

honno (approved):
LGTM!

array_api_tests/test_array_object.py
asmeurer: Oh yeah skipping is probably better.

PR #183, opened 2023-04-27
Remove use of `_UndefinedStub` in `dtype_helpers.py`

honno:
This is mighty useful as currently we have two classes of the dtype helpers—those that consume stubs and those who don't, which is hard to keep track. This unifies the approach of the helpers so they are only constructed with available dtypes.

If this was merged it'd remove the need for #181, although we might want to add automatic xfails when parametrized dtypes (i.e. in `test_iinfo`) are not found.


honno:
Folded into #184 as it goes hand-in-hand with changes there now

PR #185, opened 2023-04-27
Skip `test_matmul` and `test_remainder`

honno:
Both seem to fail the test suite sometimes, can look into later and leave skipped for now.

### DIFF
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -277,6 +277,7 @@ def test_inv(x):
 
     # TODO: Test that the result is actually the inverse
 
+@pytest.mark.skip(reason="flaky")
 @given(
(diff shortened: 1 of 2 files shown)


asmeurer:
I just did a ndindex release that might have caused this. I'll have a look.

asmeurer:
Ah yes, this did break. I plan on fixing it in #101 but need to do some other fixes first. I can make a quicker workaround fix if it becomes an issue sooner.

asmeurer:
By the way the remainder thing is some bug on the CI platform. I have to skip all these tests on CI https://github.com/data-apis/array-api-compat/blob/main/numpy-skips.txt. Would be nice to figure out what's going on there but I don't think it's really worth the time (my guess is there's some bug in the cmath library used in GitHub Actions).

PR #192, opened 2023-07-15
Fix test_take to make axis optional when ndim == 1

### LINKED ISSUE #191: Test `axis=None` in `test_take`
See https://github.com/data-apis/array-api/pull/644

asmeurer:
I didn't explicitly test axis=None because it's not clear to me that should actually be supported, given that that's the same as axis=0.

Fixes #191

### DIFF
--- a/array_api_tests/test_indexing_functions.py
+++ b/array_api_tests/test_indexing_functions.py
@@ -22,7 +22,14 @@ def test_take(x, data):
     # * negative axis
     # * negative indices
     # * different dtypes for indices
-    axis = data.draw(st.integers(0, max(x.ndim - 1, 0)), label="axis")
+
+    # axis is optional but only if x.ndim == 1
+    _axis_st = st.integers(0, max(x.ndim - 1, 0))
+    if x.ndim == 1:
+        kw = data.draw(hh.kwargs(axis=_axis_st))
+    else:
+        kw = {"axis": data.draw(_axis_st)}
+    axis = kw.get("axis", 0)
     _indices = data.draw(
         st.lists(st.integers(0, x.shape[axis] - 1), min_size=1, unique=True),
         label="_indices",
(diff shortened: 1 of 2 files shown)


honno (approved):
This LGTM, just have a few minor points that encroach on bikesheding :sweat_smile: I think I would write this something like

```python
def test_take(x, data):
    ...
    axis_strat = st.integers(0, max(x.ndim - 1, 0))
    if x.ndim == 1:  # i.e. axis is optional
        axis_strat |= st.none()
    axis = data.draw(axis_strat, label="axis")
    _axis = axis or 0
    _indices = data.draw(
        st.lists(st.integers(0, x.shape[_axis] - 1), min_size=1, unique=True),
        label="_indices",
    )
    indices = xp.asarray(_indices, dtype=dh.default_int)
    note(f"{indices=}")
    kw = data.draw(hh.specified_kwargs(("axis", axis, None)), label="kw")

    out = xp.take(x, indices, **kw)
    ph.assert_dtype("take", in_dtype=x.dtype, out_dtype=out.dtype)
    ph.assert_shape(
        "take",
        out_shape=out.shape,
        expected=x.shape[:_axis] + (len(_indices),) + x.shape[_axis + 1 :],
        kw=dict(
            x=x,
            indices=indices,
            axis=axis,
        ),
    )
    out_indices = sh.ndindex(out.shape)
    axis_indices = list(sh.axis_ndindex(x.shape, _axis))
    ...
```

* Have a core strategy that we "add" `st.none()` too if relevant—if you didn't know, piping strategies is syntactic sugar for [`st.one_of()`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.one_of).
* Generate `axis` as the actual kwarg value, and have `_axis` as the processed/explicit logical value.
* We can utilise `hh.specified_kwargs()` for when we want to test not passing the kwarg. [`test_linspace`](https://github.com/data-apis/array-api-tests/blob/f82c7bc8627cc2c3a44fa3e425f53a253a609aa8/array_api_tests/test_creation_functions.py#L476-L482) is a good example of its use.

I say all this because most of the existing tests are written to follow the above coding patterns—we don't have to be consistent or anything, just good to be aware of the prior work.

asmeurer:
I mentioned it in the OP, but I intentionally didn't test axis=None, because it's not clear to me that we actually want that keyword default, given that it's the same as axis=0 for the current semantics. Libraries like NumPy use `axis=None` to flatten the input array, and it wasn't clear that we would ever want to standardize that behavior.  This is something to bring up in the spec repo.

ev-br:
The test got merged in https://github.com/data-apis/array-api-tests/pull/385

PR #198, opened 2023-11-09
Add a test that the `newaxis` constant exists

rgommers:
The spec says that this has to be an alias to None, so verify that.


asmeurer:
Looks simple enough. We should also investigate why the has_name tests didn't catch this, but we can do that separately.

PR #203, opened 2023-11-14
Fix linalg.trace test, result dtype was incorrect

rgommers:
As the spec says, the output dtype should be the default integer/float/complex dtype just like for `sum`. Given that the reference was implemented with `sum`, avoiding to pass an explicit dtype should be enough to obtain the correct results.


honno (approved):
LGTM thanks.

Issue #190, opened 2023-06-06
test_square test failure with numpy complex64


asmeurer:
```
E                   AssertionError: out=(4+512128000j), but should be roughly x²=512128008j [square()]
E                     x=(16002+16002j)
E                   assert False
E                    +  where False = isclose(4.0, 0.0, 3.4028234663852886e+38)
E                    +    where 4.0 = (4+512128000j).real
E                    +    and   0.0 = 512128008j.real
E                   Falsifying example: test_square(
E                       x=Array(16002.+16002.j, dtype=complex64),
E                   )
```

I haven't investigated further other than to confirm that it does indeed do that

```py
>>> np.square(np.asarray(16002.+16002.j, dtype=np.complex64))
(4+512128000j)
```

I don't know if this is a bug in NumPy or just an expected numerical inaccuracy that we shouldn't b testing. I suspect the latter. Note that the relative error is around 1e-8

```py
>>> abs(512128008j - (4+512128000j))/abs(512128008j)
1.7464914572684647e-08
```

OTOH, PyTorch does give a purely imaginary answer as expected:

```py
>>> torch.square(torch.asarray(16002.+16002.j, dtype=torch.complex64))
tensor(0.+5.1213e+08j)
```

asmeurer:
Actually this failure comes up with all array libraries. It seems the test is just wrong for floating-point arithmetic.

asmeurer:
A related error with numpy

```
array_api_tests/test_operators_and_elementwise_functions.py:1403: in test_square
    unary_assert_against_refimpl(
array_api_tests/test_operators_and_elementwise_functions.py:264: in unary_assert_against_refimpl
    assert isclose(scalar_o.real, expected.real, M), msg
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = -inf, b = nan, M = 1.7976931348623157e+308

    def isclose(
        a: float,
        b: float,
        M: float,
        *,
        rel_tol: float = 0.25,
        abs_tol: float = 1,
    ) -> bool:
        """Wraps math.isclose with very generous defaults.
    
        This is useful for many floating-point operations where the spec does not
        make accuracy requirements.
        """
        if math.isnan(a) or math.isnan(b):
>           raise ValueError(f"{a=} and {b=}, but input must be non-NaN")
[...]

PR #204, opened 2023-11-14
Fix complex scalar checking and `test_square` failure

rgommers:
Closes gh-190


asmeurer:
We'll need to check if this cleans up some of the xfails in the compat library.

PR #208, opened 2023-11-15
Update to most recent array-api commit

jakevdp:


array_api_tests/dtype_helpers.py
```diff
@@ -346,6 +346,10 @@ def result_type(*dtypes: DataType):
     "boolean": (xp.bool,),
     "integer": all_int_dtypes,
     "floating-point": real_float_dtypes,
+    "real-valued": real_float_dtypes,
+    "real-valued floating-point": real_float_dtypes,
+    "complex floating-point": complex_dtypes,
+    "complex-floating point": complex_dtypes,
```
rgommers, line 352: I was initially surprised that it didn't match the strings in https://data-apis.org/array-api/latest/API_specification/generated/array_api.isdtype.html, but this is only used for text matching in `result_dtype` below. So should be correct I think.

rgommers (approved):
I'll leave this to @honno or @asmeurer to merge, but LGTM. Thanks again @jakevdp!

array_api_tests/dtype_helpers.py
jakevdp: Exactly: it's pattern matching the text in the parameter descriptions of the function docstrings. For example, `complex-floating point` [here](https://github.com/data-apis/array-api/blob/0f64005c7af6cc1906f310e929d0b8fcac24fd41/src/array_api_stubs/_2022_12/elementwise_functions.py#L730), and `complex floating-point` [here](https://github.com/data-apis/array-api/blob/0f64005c7af6cc1906f310e929d0b8fcac24fd41/src/array_api_stubs/_2022_12/elementwise_functions.py#L1233).

I wonder if there's a better way for that information to be encoded in the array api?

honno (changes requested):
Thanks for this. Unfortunately updating the submodule breaks how we generate tests for special cases (and probably other things), so will take a look myself.

array_api_tests/dtype_helpers.py
kgryte: This is a typo in this spec and should be fixed there. I will submit a patch shortly.

kgryte: A PR was [submitted](https://github.com/data-apis/array-api/pull/706) which fixes the typo. Once merged, the inclusion of `complex-floating point` will no longer be necessary, including in the 2022.12 specification, and, I believe, can be removed from this PR, provided that the submodule is updated accordingly.

honno:
Closing as addressed in https://github.com/data-apis/array-api-tests/pull/213, which uses the work here—thanks again @jakevdp and let me know if this sorts out https://github.com/data-apis/array-api/issues/631

Issue #199, opened 2023-11-10
No `default_complex` when `xp.asarray` is not present


mtsokol:
Hi!
I noticed that there isn't a default complex dtype when `xp.asarray` is not present if branch. Is it on purpose, or could `default_complex` be added there? (it causes a setup failure when running tests for a lib without `xp.asarray`)

https://github.com/data-apis/array-api-tests/blob/b5ed713ae6be7fb45cebbcbabda035b39a265187/array_api_tests/dtype_helpers.py#L237

honno:
Nice catch, yeah this would be a problem. Although generally if a library doesn't have `xp.asarray` not much is going to get tested anyway :sweat_smile:

Issue #187, opened 2023-05-25
Bump `array-api` submodule


honno:
Should resolve https://github.com/data-apis/array-api/issues/631, clear up warnings, and remove some hacks in `_array_module.py`.

Ideally we utilise a stubs package, but that doesn't exist yet (https://github.com/data-apis/array-api/pull/472). The effort to bump `array-api` could be great enough that its worth me focusing on the stubs packaging, will see.

asmeurer:
Another side-effect of this is that test_has_names currently doesn't check for any new functions added in the 2022 version of the spec.

PR #209, opened 2023-11-16
Add `default_complex` when `xp.asarray` isn't present

mtsokol:
Hi!
This PR adds `default_complex` when `xp.asarray` isn't present.

Issue: https://github.com/data-apis/array-api-tests/issues/199

### DIFF
--- a/array_api_tests/dtype_helpers.py
+++ b/array_api_tests/dtype_helpers.py
@@ -235,7 +235,12 @@ dtype_components = _make_dtype_mapping_from_names(
 if not hasattr(xp, "asarray"):
     default_int = xp.int32
     default_float = xp.float32
-    default_complex = xp.complex64 if api_version > "2021.12" else None
+    # TODO: when api_version > '2021.12', just assign to xp.complex64,
+    # otherwise default to None. Need array-api spec to be bumped first (#187).
+    try:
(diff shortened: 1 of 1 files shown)


array_api_tests/dtype_helpers.py
```diff
@@ -235,6 +235,7 @@ class MinMax(NamedTuple):
 if not hasattr(xp, "asarray"):
     default_int = xp.int32
     default_float = xp.float32
+    default_complex = xp.complex64 if api_version > "2021.12" else None
```
honno, line 238: So `xp` here actually is a wrapped array module, meaning attributes will exist even if not present in the library... except as we haven't bumped to the post-`2022.12` spec yet (#187) so this will currently give us an attribute error when testing libraries that don't have `complex64`.

Let's do something like

```suggestion
    # TODO: when api_version > '20221.12', just assign to xp.complex64,
    # otherwise default to None. Need array-api spec to be bumped first (#187).
    try:
        default_complex = xp.complex64
    except AttributeError:
        default_complex = None
    default_complex = xp.complex64 if api_version > "2021.12" else None
```

mtsokol: Sure!

honno:
Thanks :)

PR #217, opened 2023-11-24
MAINT: Add `pytest.ini` file

mtsokol:
Hi!

A small change: Due to the fact that here's no `pytest.ini`, a wrong rootdir gets picked in the downstream libs CI. E.g. in https://github.com/mtsokol/numpy/actions/runs/6982999059/job/19003296323?pr=4, there's `rootdir: /home/runner/work/numpy/numpy` instead of `${GITHUB_WORKSPACE}/array-api-tests`.

Also discussed here: https://github.com/pytest-dev/pytest/issues/11636


honno (approved):
Ah makes sense, thanks!

PR #219, opened 2023-12-05
symmetric_matrices: draw from finite if necessary

jakevdp:
Currently, running `array_api_tests/meta/test_hypothesis_helpers.py::test_symmetric_matrices` produces a warning:
```
hypothesis.errors.HypothesisWarning: bool(shared(booleans(), key='finite')) is always True, did you mean to draw a value?
```
The reason is that this test passes a boolean hypothesis parameter to the `symmetric_matrices` utility, and we must `draw` that parameter before using it. Other tests pass a raw boolean value to the function, so a reasonable fix is to only `draw` the value if it's not already a `bool` instance.

With the change in this PR, the test no longer leads to a warning.


honno (approved):
Nice catch, LGTM thanks. I just pushed a commit to simplify the test generally.

Just incase you weren't aware, we have a [`--ci` flag](https://github.com/data-apis/array-api-tests#ci-flag) for `pytest` runs, which don't run these meta tests.

PR #222, opened 2023-12-06
Test that nonzero on zero-dimensional array raises an exception.

jakevdp:
This is required by the spec. From https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.nonzero.html:

> Parameters
>     **x** (*array*) – input array. Must have a positive rank. If `x` is zero-dimensional, the function must raise an exception.

I made this a separate test so that implementations can mark it to skip if desired (this currently raises a `DeprecationWarning` in NumPy, but will error in a future release).

### DIFF
--- a/array_api_tests/test_searching_functions.py
+++ b/array_api_tests/test_searching_functions.py
@@ -87,14 +87,17 @@ def test_argmin(x, data):
         ph.assert_scalar_equals("argmin", type_=int, idx=out_idx, out=min_i, expected=expected)
 
 
+@given(xps.arrays(dtype=xps.scalar_dtypes(), shape=hh.shapes(min_dims=0, max_dims=0)))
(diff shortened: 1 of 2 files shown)


honno (approved):
Nice catch thanks! LGTM.

Historically I've tested errors all in the same function to keep with some platonic ideal of 1 ["primary test"](https://github.com/data-apis/array-api-tests#primary-tests) per function, e.g.

https://github.com/data-apis/array-api-tests/blob/d58fb6bff225af276abf0d9d130cf0f687f618ca/array_api_tests/test_manipulation_functions.py#L131-L135

which I'm not sure is worth it or that practical for either us maintaining `array-api-tests` or adopting libraries all the time. In any case, having a separate test makes sense here for sure.

jakevdp:
Thanks!

PR #227, opened 2024-01-02
Update `test_sign` for complex inputs

mtsokol:
Hi!
This PR updates `test_sign` for complex inputs to [align with Array API](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sign.html).

Relevant discussion: https://github.com/numpy/numpy/pull/25441#issuecomment-1873998082

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,6 +117,10 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
+# IDE
+.idea/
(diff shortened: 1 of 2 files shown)


honno (approved):
Thanks! I replaced use of `xp.abs` with `math.abs`, as `refimpl` only pertains to Python scalars as opposed to 0d arrays.

PR #236, opened 2024-02-26
Short-circuit with vectorisation `ph.assert_array_elements()`

honno:
And general `test_eye` improvements. Carries on the work by @jakevdp in #200 (separate PR to make things easier to track).

~Requires `signbit` released in the `2023.12` spec, which isn't even supported yet in array API adoptions like `jax.experimental.array_api`. Probably what I'll do is ignore asserting pos/neg zeros for most of the tests where it's not really relevant, so we get the vectorisation benefits regardless of API version and adoption. I'll leave that for a future PR as I'd just want to mull that over, and want to see what the shape of vectorising everything looks like first anywho. This PR is essentially a proof of concept to be applied everywhere possible.~ Let's just ignore the `signbit` stuff if not available as the immediate perf benefits for libraries probably always trumps testing signed zero stuff.

### DIFF
--- a/array_api_tests/pytest_helpers.py
+++ b/array_api_tests/pytest_helpers.py
@@ -422,25 +422,30 @@ def assert_fill(
 
 
 def _real_float_strict_equals(out: Array, expected: Array) -> bool:
-    assert hasattr(_xp, "signbit")  # sanity check
-
     nan_mask = xp.isnan(out)
     if not xp.all(nan_mask == xp.isnan(expected)):
         return False
+    ignore_mask = nan_mask
 
-    out_zero_mask = out == 0
-    out_sign_mask = xp.signbit(out)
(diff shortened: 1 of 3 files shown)


array_api_tests/pytest_helpers.py
```diff
@@ -420,6 +421,35 @@ def assert_fill(
...
 
+def _real_float_strict_equals(out: Array, expected: Array) -> bool:
+    nan_mask = xp.isnan(out)
+    if not xp.all(nan_mask == xp.isnan(expected)):
+        return False
+    ignore_mask = nan_mask
+
+    # Test sign of zeroes if xp.signbit() available, otherwise ignore as it's
+    # not that big of a deal for the perf costs.
+    if api_version >= "2023.12" and hasattr(_xp, "signbit"):
```
asmeurer, line 432: Why do we need the version check here? If the library has signbit then we can use it.

PR #237, opened 2024-02-27
Avoid passing a list of axes to tensordot

asmeurer:
cupy has logic that fails if axis is not a tuple. The standard type annotations only require axes to be a tuple.


honno (approved):
LGTM!

PR #239, opened 2024-03-01
Improvements to test_solve

asmeurer:
- Fix it to only generate arguments with mutually promotable dtypes
- Test dtype promotion
- Test output shape

Note that the standard is currently wrong for the `solve` output shape https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.solve.html#array_api.linalg.solve. It says 

> The returned array must have the same shape as x2

But this is wrong, it should have a shape like `(...,) + x2.shape`, where `(...,)` is the broadcasted stack shapes of x1 and x2.


honno (approved):
All looks good!

PR #240, opened 2024-03-07
Avoid using .size in test_eye

asmeurer:
.size doesn't work in pytorch and cannot be worked around even in the compat library.


asmeurer:
This is blocking some compat library work so I'm going to just merge this. It's a one-line fix and I manually confirmed it fixes the array_api_compat.torch test failure.

rgommers (approved):
This works I believe. In other places we use `size = math.prod(x.shape)` for this same reason.

PR #242, opened 2024-03-11
Fix `test_solve` generating arrays larger than our max array size

### LINKED ISSUE #238: test_solve - hypothesis.errors.InvalidArgument: min_size=8_325 is larger than Hypothesis is designed to handle while generating 'x2'
```
__________________________________ test_solve __________________________________

    @pytest.mark.xp_extension('linalg')
>   @given(*solve_args())

array_api_tests/test_linalg.py:616: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'ListStrategy' object has no attribute 'element_strategy'") raised in repr()] ListStrategy object at 0x7eff05717450>
elements = builds(complex, FloatStrategy(min_value=-1.8446742974197924e+19, max_value=1.8446742974197924e+19, allow_nan=False, sm...9, max_value=1.8446742974197924e+19, allow_nan=False, smallest_nonzero_magnitude=1.1754943508222875e-38).map(downcast))
min_size = 8325, max_size = 8325

    def __init__(self, elements, min_size=0, max_size=float("inf")):
        super().__init__()
        self.min_size = min_size or 0
        self.max_size = max_size if max_size is not None else float("inf")
        assert 0 <= self.min_size <= self.max_size
        if min_size > BUFFER_SIZE:
                f"min_size={min_size:_d} is larger than Hypothesis is designed to handle"
            )
E           hypothesis.errors.InvalidArgument: min_size=8_325 is larger than Hypothesis is designed to handle
E           while generating 'x2' from one_of(floating_dtypes(), complex_dtypes()).flatmap(lambda d: arrays(d, *args, elements=elements, **kwargs))
E           
E           You can reproduce this example by temporarily adding @reproduce_failure('6.98.13', b'AXicU2VlYGBkZEAHjCCoCwAFRABd') as a decorator on your test case

/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/hypothesis/strategies/_internal/collections.py:147: InvalidArgument
```

I'm not completely clear from the information given what the issue is. I don't see any issues with the arrays generated for test_solve https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/test_linalg.py#L592, nor have I seen this error before.

honno:
Resolves #238—the end side of the second `test_solve` array shape needed to be ceiled relative to the rest of the shape size, otherwise the array size could exceed our desired maximum!

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -312,8 +312,8 @@ def positive_definite_matrices(draw, dtypes=xps.floating_dtypes()):
 @composite
 def invertible_matrices(draw, dtypes=xps.floating_dtypes(), stack_shapes=shapes()):
     # For now, just generate stacks of diagonal matrices.
-    n = draw(integers(0, SQRT_MAX_ARRAY_SIZE),)
     stack_shape = draw(stack_shapes)
+    n = draw(integers(0, SQRT_MAX_ARRAY_SIZE // max(math.prod(stack_shape), 1)),)
     dtype = draw(dtypes)
     elements = one_of(
         from_dtype(dtype, min_value=0.5, allow_nan=False, allow_infinity=False),
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -19,6 +19,7 @@ from hypothesis.strategies import (booleans, composite, tuples, floats,
                                    data)
 from ndindex import iter_indices
 
+import math
 import itertools
 from typing import Tuple
 
@@ -610,8 +611,9 @@ def solve_args() -> Tuple[SearchStrategy[Array], SearchStrategy[Array]]:
 
     @composite
     def _x2_shapes(draw):
-        end = draw(integers(0, SQRT_MAX_ARRAY_SIZE))
-        return draw(stack_shapes)[1] + draw(x1).shape[-1:] + (end,)
+        base_shape = draw(stack_shapes)[1] + draw(x1).shape[-1:]
+        end = draw(integers(0, SQRT_MAX_ARRAY_SIZE // max(math.prod(base_shape), 1)))
+        return base_shape + (end,)
 
     x2_shapes = one_of(x1.map(lambda x: (x.shape[-1],)), _x2_shapes())
     x2 = arrays(shape=x2_shapes, dtype=mutual_dtypes.map(lambda pair: pair[1]))
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -20,11 +20,12 @@ from hypothesis.strategies import (booleans, composite, tuples, floats,
 from ndindex import iter_indices
 
 import itertools
+from typing import Tuple
 
 from .array_helpers import assert_exactly_equal, asarray
 from .hypothesis_helpers import (arrays, all_floating_dtypes, xps, shapes,
                                  kwargs, matrix_shapes, square_matrix_shapes,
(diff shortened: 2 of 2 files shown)


asmeurer:
The change here seems correct, though maybe we should get it fully working before merging. 

If the source of this error is passing too large of a shape to `arrays` perhaps hypothesis should have a separate check for this in `arrays`, or else intercept the error there and translate it to a more meaningful one.

Zac-HD:
> If the source of this error is passing too large of a shape to `arrays` perhaps hypothesis should have a separate check for this in `arrays`, or else intercept the error there and translate it to a more meaningful one.

So far as I can tell, this strategy does not go through `arrays()`.

asmeurer:
There's this line https://github.com/data-apis/array-api-tests/pull/242/files#diff-6056c0b3af9cd3ba66387432a17f5f36bbd54220419656441a8b01bcdc4df44bR619. 

Also, I'll need to check again tomorrow, but from what I recall from a week ago when I ran this through a debugger there was some line in hypothesis in the `arrays` code that looked like it was what was generating the long list (but I also found it very difficult to tell what was going on inside of the hypothesis stack in the debugger).

honno:
> > If the source of this error is passing too large of a shape to `arrays` perhaps hypothesis should have a separate check for this in `arrays`, or else intercept the error there and translate it to a more meaningful one.
> 
> So far as I can tell, this strategy does not go through `arrays()`.

Looks like our `hh.invertible_matrices()` strategy does use `arrays()` and like before there seemed to be a case where the final array shape could exceed our max size, so pushed another fix if you could take a gander @asmeurer.

> `ARRAY_API_TESTS_MODULE=numpy.array_api pytest array_api_tests/test_linalg.py -k solve --max-examples=1000`

On my end I get a failed deadline health check for this, both before and after my changes :thinking: When ignoring the deadline (`from hypothesis import settings; @settings(deadline=None)`) the test passes locally both before and after my changes too.

asmeurer:
Found the line in hypothesis I was thinking about. It's actually in the traceback in the gist I posted 

```
    |   File "/Users/aaronmeurer/miniconda3/envs/array-apis/lib/python3.11/site-packages/hypothesis/extra/array_api.py", line 369, in do_draw
    |     elems = data.draw(
    |             ^^^^^^^^^^

```

If I am reading this correctly, `arrays` does indeed use `lists` https://github.com/HypothesisWorks/hypothesis/blob/5eeb51ad38a44fefdc4b82dc0062487968ee089a/hypothesis-python/src/hypothesis/extra/array_api.py#L370

asmeurer:
> On my end I get a failed deadline health check for this, both before and after my changes 🤔 When ignoring the deadline (from hypothesis import settings; @settings(deadline=None)) the test passes locally both before and after my changes too.

I don't know. Maybe it requires a lot of examples. Indeed if `invertible_matrices` alone can cause this I've never seen it in other tests that use it. I just ran a whole bunch of examples for test_solve and I don't see this any more in this branch. 

By the way, in the test suite you can disable the deadline with `--disable-deadline`.

honno:
> I just ran a whole bunch of examples for test_solve and I don't see this any more in this branch.

So in any case, if you're not getting a failed health check now then you/I can merge this? And I'll mull over the other points :sweat_smile:

asmeurer:
I did get some deadline errors, but I ignored them.

PR #251, opened 2024-04-08
Added check for numpy version as np.bool is deprecated

fnhirwa:
Numpy deprecated the use of `np.bool` since `version: 1.20.0` 
https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

With the current version of Numpy when I ran the tests I was getting the following errors

By setting numpy as array_module: os.environ["ARRAY_API_TESTS_MODULE"] = "numpy"
image

The reason for this is that some frameworks like [ivy](https://github.com/unifyai/ivy) implement the `array-api` functions using numpy as core module, I understand that numpy added an extension for the Array-api interface via `numpy.array_api` (still experimental) but this might be worth supporting for such frameworks implementing Array API using numpy.

### DIFF
--- a/array_api_tests/__init__.py
+++ b/array_api_tests/__init__.py
@@ -9,7 +9,6 @@ from . import _version
 
 __all__ = ["xp", "api_version", "xps"]
 
-os.environ["ARRAY_API_TESTS_MODULE"] = "numpy"
 
 # You can comment the following out and instead import the specific array module
 # you want to test, e.g. `import numpy.array_api as xp`.
--- a/array_api_tests/__init__.py
+++ b/array_api_tests/__init__.py
@@ -9,6 +9,7 @@ from . import _version
 
 __all__ = ["xp", "api_version", "xps"]
 
+os.environ["ARRAY_API_TESTS_MODULE"] = "numpy"
 
 # You can comment the following out and instead import the specific array module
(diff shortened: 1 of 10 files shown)


rgommers:
Note that in numpy 2.0.0 the main numpy namespace is array API standard compliant, including `np.bool`.

asmeurer:
This sort of fix is still useful. The test suite is currently unusable with the latest versions of numpy or cupy.

honno:
Thanks for the PR @fnhirwa! I agree with you and Aaron that it'd be nice for NumPy and CuPy to work out-of-the-box—this shouldn't be an issue for future versions but would be nice to test older versions still.

I like the approach here of having a separated boolean variable from `xp.bool`, but unfortunately there are still Hypothesis internals to contend with when you don't have `xp.bool`. So instead I opened #252 which patches the top-level `xp` module's bool, which is a somewhat less honest approach but feels more appropriate, lest we start monkey patching Hypothesis[^1]. Given support for latest NumPy/CuPy isn't that critical, I'd rather keep the maintenance overhead minimal too.

[^1]: If we did want to go that route (that can of worms was opened with `hh.arrays()` and `st.floats()`), one thing I'd prob prefer is just having a `dtype_helpers.bool` object, rather than a function we call for each test module.

PR #254, opened 2024-04-19
Reset the max-examples default back to 100

asmeurer:
The tests run fast enough now that there's no need to lower this to 20.

See the discussion at #246.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,7 +1,6 @@
 from functools import lru_cache
 from pathlib import Path
 import argparse
-import math
 import warnings
 import os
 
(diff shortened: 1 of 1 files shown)


honno:
Could we just change our own workflow to have `--hypothesis-max-examples=100`, instead of the default? As defo for something like JAX this makes it basically unusable on default runs :sweat_smile:

asmeurer:
JAX is the exception, not us. If JAX is slower than most libraries, they should manually lower their max examples. The library defaults should match the common case, not the least common denominator.

asmeurer:
I also changed the unvectorized max examples to scale linearly with max-examples instead of logarithmically. Otherwise, it's very difficult to actually run a lot of examples for unvectorized tests. We can adjust the scaling factor if 1/10 is not small enough by default.

PR #255, opened 2024-04-30
2023.12 coverage

### LINKED ISSUE #249: 2023.12 support
We need a tracking issue for 2023.12 support in the test suite. 

Here is the list of changes https://data-apis.org/array-api/latest/changelog.html#v2023-12

**New Functions**

- [x] `__array_namespace_info__` (i.e., the [inspection API](https://data-apis.org/array-api/latest/API_specification/inspection.html))
- [x] `clip`
- [x] `copysign`
- [x] `cumulative_sum`
- [x] `hypot`
- [x] `maximum`
- [x] `minimum`
- [x] `moveaxis`
- [x] `repeat`
- [x] `searchsorted`
- [x] `signbit`
- [x] `tile`
- [x] `unstack`

**API Changes**

- [ ] `device` keyword added to `astype`
- [ ] `from_dlpack` requires exceptions in some cases
- [ ] `copy` and `device` keywords to `from_dlpack`
- [ ] `sum`, `prod`, `trace` output dtype change (https://github.com/data-apis/array-api-tests/issues/234)

**Other Changes**
(not sure if these need tests)

- [ ] Support for lazy functions in various APIs. Will probably need a real world lazy implementation to support this.
- [ ] Order of signed zeros unspecified for `max` and `min`. Just need to make sure we aren't testing this. 
- [ ] Out of bounds unspecified for `take`.
- [ ] `tensordot` should support negative axes (not sure if this is already tested)

I think that's everything but you might want to read through the changes too to check if I missed something.

honno:
Should resolve #249, currently just a dump of all the testing which I know should be accurate.

### DIFF
--- a/array_api_tests/test_inspection_functions.py
+++ b/array_api_tests/test_inspection_functions.py
@@ -1,0 +1,43 @@
+import pytest
+from hypothesis import given, strategies as st
+
+from . import xp
+
+pytestmark = pytest.mark.min_version("2023.12")
+
+
+def test_array_namespace_info():
+    out = xp.__array_namespace_info__()
+
+    capabilities = out.capabilities()
+    assert isinstance(capabilities, dict)
+
+    out.default_device()
+
+    default_dtypes = out.default_dtypes()
+    assert isinstance(default_dtypes, dict)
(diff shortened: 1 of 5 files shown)


honno:
Lets get this in and see about filling in gaps after.

PR #257, opened 2024-05-01
Add a sanity check that signbit works

asmeurer:
The array-api-strict package disables signbit at runtime when the API version is set to 2022.12. This happens when the function is called, so the hasattr check would pass even though the function call fails.

### DIFF
--- a/array_api_tests/pytest_helpers.py
+++ b/array_api_tests/pytest_helpers.py
@@ -421,7 +421,8 @@ def assert_fill(
         assert xp.all(xp.equal(out, xp.asarray(fill_value, dtype=dtype))), msg
 
 
-def _has_signbit() -> bool:
+def _has_functional_signbit() -> bool:
+    # signbit can be available but not implemented (e.g., in array-api-strict)
     if not hasattr(_xp, "signbit"):
         return False
     try:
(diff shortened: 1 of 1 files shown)


array_api_tests/pytest_helpers.py
```diff
@@ -421,6 +421,15 @@ def assert_fill(
         assert xp.all(xp.equal(out, xp.asarray(fill_value, dtype=dtype))), msg
 
 
+def _has_signbit() -> bool:
```
honno, line 424: A small comment/docstring explaining that this function exists because `signbit` can be available but not implemented would help I think. Possibly worth re-naming the function too/instead, something like `_has_functional_signbit()`.

array_api_tests/pytest_helpers.py
```diff
@@ -429,7 +438,7 @@ def _real_float_strict_equals(out: Array, expected: Array) -> bool:
 
     # Test sign of zeroes if xp.signbit() available, otherwise ignore as it's
     # not that big of a deal for the perf costs.
-    if hasattr(_xp, "signbit"):
+    if _has_signbit():
```
asmeurer, line 441:
```suggestion
    if _has_functional_signbit():
```

PR #258, opened 2024-05-02
Add a header to the tests indicating the module and API version

asmeurer:


honno (approved):
LGTM, certainly useful—never played around with headers before myself but it really is that simple huh :sweat_smile:

PR #262, opened 2024-05-14
Fixes for 2023.12 tests

asmeurer:


asmeurer:
Functions to prioritize:

- `unstack` (new in NumPy)
- `clip` (has a lot of corner cases)

mtsokol:
I've got one question: Inspection functions shouldn't be in the main namespace but only in the `__array_namespace_info__` namespace, as described in the standard (https://data-apis.org/array-api/latest/API_specification/inspection.html#inspection-apis), right?

asmeurer:
Yes, that's right.

array_api_tests/test_operators_and_elementwise_functions.py
```diff
@@ -927,7 +927,7 @@ def test_ceil(x):
 
 
 @pytest.mark.min_version("2023.12")
-@given(x=hh.arrays(dtype=hh.real_floating_dtypes, shape=hh.shapes()), data=st.data())
+@given(x=hh.arrays(dtype=hh.int_dtypes, shape=hh.shapes()), data=st.data())
```
hameerabbasi, line 930: Shouldn't this be every real numerical dtype?

asmeurer:
We still need a lot more test updates for 2023, but I'm going to merge this to make it easier for people to use.

PR #267, opened 2024-05-22
Fix the signature test for outer when it isn't inspectable

asmeurer:


rgommers (approved):
LGTM, thanks @asmeurer

PR #279, opened 2024-06-27
Revert "Suppress the data_too_large health check"

asmeurer:
Reverts data-apis/array-api-tests#278


asmeurer:
Hmm, some other test failed.

asmeurer:
I think the test failure is just that we can get unlucky in the meta tests sometimes. Not going to worry about it unless it crops up again.

PR #283, opened 2024-07-30
Fix way to determine default_complex

cbourjau:
The current test suite crashes during the collection time if the implementation in question does not support complex data types. This PR changes the way in which the `default_complex` type is determined to avoid the issue. Further discussion and context can be found [here](https://github.com/Quantco/ndonnx/pull/32).

### DIFF
--- a/array_api_tests/dtype_helpers.py
+++ b/array_api_tests/dtype_helpers.py
@@ -322,15 +322,16 @@ else:
     default_float = xp.asarray(float()).dtype
     if default_float not in real_float_dtypes:
         warn(f"inferred default float is {default_float!r}, which is not a float")
-    if api_version > "2021.12":
+
+    if not (hasattr(xp, "complex32") or hasattr(xp, "complex64")):
+        default_complex = None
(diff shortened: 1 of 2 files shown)


array_api_tests/dtype_helpers.py
```diff
@@ -322,15 +322,16 @@ def accumulation_result_dtype(x_dtype, dtype_kwarg):
     default_float = xp.asarray(float()).dtype
     if default_float not in real_float_dtypes:
         warn(f"inferred default float is {default_float!r}, which is not a float")
-    if api_version > "2021.12":
```
asmeurer, line 325: I don't know about removing this. This is the check that's currently done throughout the suite to check for complex support. 

Maybe we should instead make sure that `ARRAY_API_TESTS_SKIP_DTYPES` works correctly for complex dtypes.

cbourjau: I readded the version check and added a check that no complex types are requested to be skipped.

PR #286, opened 2024-08-08
Remove `array-api-tests` prefix in ID matching

jakevdp:
This fixes failures I'm seeing when attempting to update array-api-tests in https://github.com/google/jax/pull/22903

Since the start of each skipped test name must be `array_api_tests`, I think this is always safe: https://github.com/data-apis/array-api-tests/blob/31468af33199d43b2d90f207bdffa68bfda47470/conftest.py#L159-L161

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -124,6 +124,8 @@ def xp_has_ext(ext: str) -> bool:
 
 
 def check_id_match(id_, pattern):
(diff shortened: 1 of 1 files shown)


asmeurer:
I don't quite understand why you're seeing this error, although I have seen in the past that it sometimes wants to prepend some other directory to the test name. I agree just removing it after parsing should be fine.

Issue #230, opened 2024-01-17
Deprecation warnings from numpy fft 


asmeurer:
The numpy.array_api has several test failures due to https://github.com/data-apis/array-api/pull/720. But additionally, there are also these warnings, which should be looked into:

```
array_api_tests/test_fft.py::test_fftn
  /Users/aaronmeurer/Documents/numpy/numpy/fft/_pocketfft.py:857: DeprecationWarning: `axes` should not be `None` if `s` is not `None` (Deprecated in NumPy 2.0). In a future version of NumPy, this will raise an error and `s[i]` will correspond to the size along the transformed axis specified by `axes[i]`. To retain current behaviour, pass a sequence [0, ..., k-1] to `axes` for an array of dimension k.
    return _raw_fftnd(a, s, axes, fft, norm)

array_api_tests/test_fft.py: 16 warnings
  /Users/aaronmeurer/Documents/numpy/numpy/array_api/fft.py:190: DeprecationWarning: `axes` should not be `None` if `s` is not `None` (Deprecated in NumPy 2.0). In a future version of NumPy, this will raise an error and `s[i]` will correspond to the size along the transformed axis specified by `axes[i]`. To retain current behaviour, pass a sequence [0, ..., k-1] to `axes` for an array of dimension k.
    res = Array._new(np.fft.irfftn(x._array, s=s, axes=axes, norm=norm))
```

We should also double check if the standard is correct here.

PR #297, opened 2024-11-13
MAINT: avoid axis=None and s is not None in test_{fftn,ifftn}

ev-br:
closes https://github.com/data-apis/array-api-tests/issues/230

### DIFF
--- a/array_api_tests/test_fft.py
+++ b/array_api_tests/test_fft.py
@@ -57,6 +57,9 @@ def draw_s_axes_norm_kwargs(x: Array, data: st.DataObject, *, size_gt_1=False) -
         s_strat = st.none() | s_strat
     s = data.draw(s_strat, label="s")
 
+    # Using `axes is None and s is not None` is disallowed by the spec
+    assume(axes is not None or s is None)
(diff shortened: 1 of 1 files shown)


asmeurer:
Indeed, this was fixed in the standard (see https://github.com/data-apis/array-api/pull/720/files#diff-4333927b52a89245b056bfb25bd59e365c8c48ad92153ece1fb5974f0f911fffR144).

array_api_tests/test_fft.py
```diff
@@ -57,6 +57,9 @@ def draw_s_axes_norm_kwargs(x: Array, data: st.DataObject, *, size_gt_1=False) -
         s_strat = st.none() | s_strat
     s = data.draw(s_strat, label="s")
 
+    if axes is None and s is not None:
```
asmeurer, line 60: It's better to just tell hypothesis to skip this case, like

```
assume(axis is not None or s is not None)
```

ev-br: Nice, thanks! Updated to the suggestion.

Issue #235, opened 2024-02-20
Update libraries we test on CI


honno:
- Firstly we really need to
  - [ ] Replace `numpy.array_api` with [`array-api-strict`](https://github.com/data-apis/array-api-strict)
    - `numpy.array_api` is going to be removed and supplanted by this third-party library
    - As bought up in https://github.com/data-apis/array-api-tests/pull/101#issuecomment-1949382387
  - [ ] Also cover `meta/` tests (we never did before!)
    - Would require one or two fixes of those tests. 
- Then ideally we're testing libraries covered by [`array-api-compat`](https://github.com/data-apis/array-api-compat)
  - [ ] NumPy
  - [ ] PyTorch
  - [ ] Dask
  - CuPy is probably a no-go because GPUs on CI is too finnicky
- I think it'd be really nice to test `jax.experimental.array_api` too as it has some distinct properties that we want to be mindful with for future test suite development.

asmeurer:
For array-api-compat I have a script that I can run on a CUDA machine that runs the CuPy tests. We could have something like that here. We could also do that for PyTorch GPU. We would then just have to run it ourselves manually from time to time.

PR #298, opened 2024-11-14
CI: parametrize over Array API modules

ev-br:
towards https://github.com/data-apis/array-api-tests/issues/235

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,8 @@ jobs:
         elif [[ ${{ matrix.array-api-module }} == *"jax.numpy"* ]]
         then
           python -m pip install "jax[cpu]"
+        elif [[ ${{ matrix.array-api-module }} == *"numpy"* ]]
+          python -m pip install numpy
         else
           python -m pip install ${{ matrix.array-api-module }}
         fi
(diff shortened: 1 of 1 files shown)


asmeurer:
The thing we really want to test here is that the tests themselves don't break whenever we update them or write new ones. This is especially important if we want to do some more serious refactorings here. The test suite is a little odd because we often will "break" upstream by adding a test that fails, because the upstream library is not conformant but the thing was just never tested before. But we don't want to "break" upstream by adding a test that's actually wrong, or testing something in a way that isn't really actually required by the standard (an example would be a floating-point closeness test that is too strict).

So when it comes down to it, what should go on CI here really just comes down to developer convenience. The test suite will be run upstream eventually anyways, by the upstreams themselves (and compat is probably the primary upstream since it tests 4 different libraries). On the one hand, testing more libraries here will catch more things. But on the other hand, we have to maintain a set of skips/xfails here to effectively do that, because many failures are expected. This is the case even for compat. We could reuse the xfail files in the compat repo, although that would effectively amount to just duplicating the CI here from the compat repo. The benefit of that would be catching failures faster. 

But also consider that this repo doesn't have releases. If something breaks, you can very easily revert it. 

The strict library is nice because it has almost no xfails. Plan NumPy 2.1 is like that as well, although that could change whenever new functions get added to the standard. 

For now, I've made heavy use of local testing against various libraries when doing development here. If you feel that having that on CI would be better, then you should set it up. It really comes down to what's the most convenient. That would mean either maintaining a set of xfails, or just ignoring failures entirely and just using the CI as a vehicle to run the tests. You definitely should check any updated test against the major libraries one way or the other.

PR #305, opened 2024-11-16
TST: filter out {invalid value, overflow, divide by zero} warnings

ev-br:
This is a minor usability tweak: IIUC (and please correct me if not), we are testing elementwise functions with various invalid arguments (nans, divide-by-zero, overflows) against "reference implementations". These operations generate walls of warnings, which are annoying if benign.
So mass-filter these warnings.
I suppose warning filters should be very specific to avoid masking genuine problems, so am adding the "message" argument to the filters.
In principle, one can make it even more specific if a bit more verbose--- and list warning-emitting functions explicitly: instead of

```
ignore:invalid value encountered in:RuntimeWarning
```

use

```
ignore:invalid value encountered in divide:RuntimeWarning
ignore:invalid value encountered in sqrt:RuntimeWarning
```
etc

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,7 @@
+[pytest]
+filterwarnings =
+    ignore:invalid value encountered in:RuntimeWarning
+    ignore:overflow encountered in:RuntimeWarning
+    ignore:divide by zero encountered in:RuntimeWarning
+
+
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
 filterwarnings =
+    # Ignore floating-point warnings from NumPy
     ignore:invalid value encountered in:RuntimeWarning
     ignore:overflow encountered in:RuntimeWarning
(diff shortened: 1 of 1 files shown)


asmeurer:
I think the idea in the past that this was too NumPy specific, but I tend to agree that the wall of warnings is annoying, and I don't see any particular issue with just ignoring them. I agree with keeping the warning text specific so we don't catch too much. 

Maybe if it were easier to suppress the warnings from the NumPy side this would be unnecessary, but that isn't straightforward to do from the way the test suite works as far as I can tell. I think NumPy would need something like support for setting the error state with an environment variable.

pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+filterwarnings =
+    ignore:invalid value encountered in:RuntimeWarning
```
asmeurer, line 3:
```suggestion
    # Ignore floating-point warnings from NumPy
    ignore:invalid value encountered in:RuntimeWarning
```

Issue #301, opened 2024-11-14
Incomplete tests


asmeurer:
This is a tracking issue for tests that are currently incomplete. These are mostly notated in the code by TODO comments, but it's useful to note them all here too. I haven't checked everything throughly so if there is something missing that isn't noted in a comment it might not be listed here.

Note there are also some functions that aren't yet tested at all, but those are covered in separate issues (e.g., #248 #299). I also am not including tests that are currently skipped entirely as flaky (https://github.com/data-apis/array-api-tests/issues/300).

**test_manipulation_functions.py**

- `test_tile` (no shapes or values testing)

**test_operators_and_elementwise_functions.py**

- `test_divide` (no complex numbers testing)
- `test_negative` (there is a TODO about uints)

**test_statistical_functions.py**

See also https://github.com/data-apis/array-api-tests/issues/167

- `test_mean` (no values testing)
- `test_std` (no values testing)
- `test_var` (no values testing)

**NOTE: THIS LIST IS INCOMPLETE**

ev-br:
> test_negative (there is a TODO about uints)

The TODO is `# TODO: clarify if uints are acceptable, adjust accordingly` and apparently negative(uint) wraps around in at least several libraries:

```
In [1]: import torch

In [2]: torch.negative
Out[2]: <function torch._VariableFunctionsClass.negative>

In [3]: torch.negative(torch.as_tensor(3, dtype=torch.uint8))
Out[3]: tensor(253, dtype=torch.uint8)

In [4]: import jax.numpy as jnp

In [5]: jnp.negative(jnp.asarray(3, dtype=jnp.uint8))
Out[5]: Array(253, dtype=uint8)

In [6]: import numpy as np

In [7]: np.negative(np.asarray(3, dtype=np.uint8))
Out[7]: np.uint8(253)

In [8]: np.__version__
Out[8]: '2.1.0'

In [9]: import array_api_strict as xp

In [10]: xp.negative(xp.asarray(3, dtype=xp.uint8))
Out[10]: Array(253, dtype=array_api_strict.uint8)

```

The spec does not seem to fix the behavior for unsigned ints: https://data-apis.org/array-api/latest/API_specification/generated/array_api.negative.html

So maybe a first action item is to clarify the spec?

asmeurer:
[...]

PR #306, opened 2024-11-16
stop skipping test_divide for complex inputs

ev-br:
Address the `test_divide (no complex numbers testing)` part of https://github.com/data-apis/array-api-tests/issues/301

Locally tests seem to pass, so I actually wonder what was meant by the TODO comment.


asmeurer:
Not sure. It was added in ff865bcd. It's possible my fix at aba096e6 using `cmath` instead of `math` was all that was needed.

PR #308, opened 2024-11-17
Use `math.prod`

ev-br:
Minor housekeeping. Requiring python >= 3.8 seems reasonably safe now.


asmeurer:
Indeed there should be no issues with using it now. You do sometimes do have to be a little careful that 0-D arrays (or numpy scalars) don't sneak in. But if that happens, generally it means you should have put a call to `int` somewhere.

PR #309, opened 2024-11-18
switch back testing of irfftn shapes

ev-br:
I *believe* this is a correct interpretation of https://numpy.org/doc/stable/reference/generated/numpy.fft.irfftn.html

cross-ref gh-301

### DIFF
--- a/array_api_tests/test_fft.py
+++ b/array_api_tests/test_fft.py
@@ -227,19 +227,17 @@ def test_irfftn(x, data):
         expected=dh.dtype_components[x.dtype],
     )
 
-    # TODO: assert shape correctly
-    # _axes = sh.normalize_axis(axes, x.ndim)
-    # _s = x.shape if s is None else s
-    # expected = []
-    # for i in range(x.ndim):
-    #     if i in _axes:
-    #         side = _s[_axes.index(i)]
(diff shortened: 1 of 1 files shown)


asmeurer:
Good to always check the standard itself (because sometimes there are typos in the standard, and sometimes it intentionally deviates from numpy) https://data-apis.org/array-api/latest/extensions/generated/array_api.fft.irfftn.html#irfftn. In this case, I think it also matches.

asmeurer:
But maybe you could double check yourself if it looks like the standard agrees with this. If so, let's merge this. If not, most likely we need to issue a fix to the standard (for fft it was mostly intended to match numpy; I already made a lot of fixes at https://github.com/data-apis/array-api/pull/720, but I wouldn't be surprised if I missed something)

ev-br:
Double-checked the text of the standard, and it does look like it agrees.

At some point we might want to add a test that `irfftn(rfftn(x)) == x` up to numerical accuracy, as stipulated in the standard. However this should be done across the whole fft extension (`fft(ifft(x)) == x` etc if at all---this is value testing which is probably best left to the array libraries themselves. 

Merging, thanks Aaron for the review!

Issue #289, opened 2024-09-03
Generate NumPy scalars for test inputs


asmeurer:
We should update the `arrays` strategy so that it can generate NumPy scalars whenever it would generate a 0-D array. This can be done by doing something like

```py
if a.ndim == 0 and draw(booleans()):
    a = a[()]
```

which would be a no-op in any other library. 

There are several issues with NumPy scalars and the array API that are not being caught by the test suite because they aren't being tested. See for instance https://github.com/data-apis/array-api-tests/issues/287 (NumPy scalars do not have in-place methods), and https://github.com/data-apis/array-api-compat/pull/165 (NumPy scalars were missing `__array_namespace__`, which to be fair isn't tested anyways #248).

PR #310, opened 2024-11-18
ENH: generate numpy scalars or 0D arrays

ev-br:
closes https://github.com/data-apis/array-api-tests/issues/289

### DIFF
--- a/array_api_tests/test_creation_functions.py
+++ b/array_api_tests/test_creation_functions.py
@@ -263,7 +263,7 @@ def scalar_eq(s1: Scalar, s2: Scalar) -> bool:
     data=st.data(),
 )
 def test_asarray_arrays(shape, dtypes, data):
-    x = data.draw(hh.arrays(dtype=dtypes.input_dtype, shape=shape), label="x")
+    x = data.draw(hh.arrays_no_scalars(dtype=dtypes.input_dtype, shape=shape), label="x")
     dtypes_strat = st.just(dtypes.input_dtype)
     if dtypes.input_dtype == dtypes.result_dtype:
         dtypes_strat |= st.none()
--- a/array_api_tests/test_creation_functions.py
+++ b/array_api_tests/test_creation_functions.py
@@ -263,6 +263,7 @@ def scalar_eq(s1: Scalar, s2: Scalar) -> bool:
     data=st.data(),
 )
 def test_asarray_arrays(shape, dtypes, data):
+    # generate arrays only since we draw the copy= kwd below (and np.asarray(scalar, copy=False) error out)
     x = data.draw(hh.arrays_no_scalars(dtype=dtypes.input_dtype, shape=shape), label="x")
     dtypes_strat = st.just(dtypes.input_dtype)
     if dtypes.input_dtype == dtypes.result_dtype:
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -64,7 +64,7 @@ def from_dtype(dtype, **kwargs) -> SearchStrategy[Scalar]:
 
 
 @wraps(xps.arrays)
-def arrays(dtype, *args, elements=None, **kwargs) -> SearchStrategy[Array]:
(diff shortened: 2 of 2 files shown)


asmeurer:
Looks good. Can you check the tests with numpy itself, or at least array_api_compat.numpy, and then merge this? It's possible this could uncover some test failures, and we'd want to at least be aware of them if not fix them before merging here. 

This is an example of where having numpy tests run on CI could be useful (https://github.com/data-apis/array-api-tests/pull/298). array-api-strict does not use scalars (it internally converts any numpy scalar into a 0-D array), so the CI tests for it won't really say much here.

ev-br:
It's interesting, actually. 

With numpy, `$ ARRAY_API_TESTS_MODULE=numpy pytest array_api_tests -v` shows exactly the same failures locally on main and on the PR:

```
======================================================================================== short test summary info =========================================================================================
FAILED array_api_tests/test_creation_functions.py::test_asarray_arrays - ValueError: Unable to avoid copy while creating an array as requested.
FAILED array_api_tests/test_data_type_functions.py::test_finfo[float32] - AssertionError: type(out.eps)=<class 'numpy.float32'>, but should be float [finfo(float32)]
FAILED array_api_tests/test_manipulation_functions.py::test_repeat - TypeError: Cannot cast array data from dtype('uint64') to dtype('int64') according to the rule 'safe'
FAILED array_api_tests/test_operators_and_elementwise_functions.py::test_clip - AssertionError: out.dtype=uint16, but should be uint8 [clip(uint8)]
FAILED array_api_tests/test_signatures.py::test_func_signature[argsort] - AssertionError: Argument 'descending' missing from signature
FAILED array_api_tests/test_signatures.py::test_func_signature[sort] - AssertionError: Argument 'descending' missing from signature
FAILED array_api_tests/test_signatures.py::test_func_signature[clip] - AssertionError: Expected argument 'a_min' to be named 'min'
FAILED array_api_tests/test_sorting_functions.py::test_argsort - TypeError: argsort() got an unexpected keyword argument 'descending'
FAILED array_api_tests/test_sorting_functions.py::test_sort - TypeError: sort() got an unexpected keyword argument 'descending'
======================================================================== 9 failed, 989 passed, 16 skipped, 457 warnings in 50.98s 
```

with `array_api_compat.numpy` though, a small additional fix is needed to avoid

```
(Pdb) import numpy as np
(Pdb) p np.__version__
'2.1.3'
(Pdb) p np.asarray(np.False_, copy=False)
*** ValueError: Unable to avoid copy while creating an array as requested.
If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.
```

So, two things to discuss:

- could you take a look that https://github.com/data-apis/array-api-tests/pull/310/commits/0f59c216262f23f33e1d59968b19d71899c1d855 makes sense?
- how come it's not caught by testing with just numpy, neither locally nor on CI. Just too few examples? testing with `$ ARRAY_API_TESTS_MODULE=array_api_compat.numpy pytest array_api_tests -v` catches it consistently from the first try.

array_api_tests/test_creation_functions.py
```diff
@@ -263,7 +263,7 @@ def scalar_eq(s1: Scalar, s2: Scalar) -> bool:
     data=st.data(),
 )
 def test_asarray_arrays(shape, dtypes, data):
-    x = data.draw(hh.arrays(dtype=dtypes.input_dtype, shape=shape), label="x")
+    x = data.draw(hh.arrays_no_scalars(dtype=dtypes.input_dtype, shape=shape), label="x")
```
ev-br, line 266:
```suggestion
    # generate arrays only since we draw the copy= kwd below (and np.asarray(scalar, copy=False) error out)
    x = data.draw(hh.arrays_no_scalars(dtype=dtypes.input_dtype, shape=shape), label="x")
```

ev-br:
Merging after an offline discussion with Aaron.

PR #313, opened 2024-11-20
Avoid calling hh.arrays with dtype=None for complex types

cbourjau:
### Issue description

[ndonnx](https://github.com/Quantco/ndonnx) does not (yet) support complex data types. We signal that by setting `ARRAY_API_TESTS_SKIP_DTYPES="complex64,complex128"`. During the collection phase [`hypothesis_helpers.complex_dtypes`](https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/hypothesis_helpers.py#L190) is set to `None`. However, downstream calls to `hypothesis_helpers.arrays` appear to not be able to handle `dtype=None` as an argument, ultimately leading to an error when looking up the data type such as:
```
________________ ERROR collecting array_api_tests/test_operators_and_elementwise_functions.py ________________________
api-coverage-tests/array_api_tests/test_operators_and_elementwise_functions.py:1066: in <module>
    @given(hh.arrays(dtype=hh.complex_dtypes, shape=hh.shapes()))
api-coverage-tests/array_api_tests/hypothesis_helpers.py:73: in arrays
    elements = from_dtype(dtype)
api-coverage-tests/array_api_tests/hypothesis_helpers.py:44: in from_dtype
    min_, max_ = dh.dtype_ranges[component_dtype]
api-coverage-tests/array_api_tests/dtype_helpers.py:85: in __getitem__
    raise KeyError(f"{key!r} not found")
E   KeyError: 'None not found'

```
Since data types are looked up using `__eq__` things appear to work if any of the data types compare equal to `None` which is the case for `numpy.dtype("float64")`, for example.

### Solution
This PR simply guards each affected tests (i.e. tests that work on `hypothesis_helpers.complex_dtypes`) with a respective if statement. This works just fine in practice for us, but I'm sure there is a better fix out there. I tried to set `complex_dtypes = ()`, which seems reasonable, but that lead to other errors.

### DIFF
--- a/array_api_tests/hypothesis_helpers.py
+++ b/array_api_tests/hypothesis_helpers.py
@@ -10,7 +10,7 @@ from typing import Any, List, Mapping, NamedTuple, Optional, Sequence, Tuple, Un
 from hypothesis import assume, reject
 from hypothesis.strategies import (SearchStrategy, booleans, composite, floats,
                                    integers, just, lists, none, one_of,
-                                   sampled_from, shared, builds)
+                                   sampled_from, shared, builds, nothing)
 
 from . import _array_module as xp, api_version
 from . import array_helpers as ah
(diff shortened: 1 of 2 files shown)


asmeurer:
Does using pytest.skip not work here? That would be preferred if it can work.

asmeurer:
You could also try modifying hh.arrays so that it fails at example draw time instead of at function call time when dtypes=().

I'm curious how this is able to work when complex dtypes aren't enabled due to older standard versions. Maybe that isn't ever actually tested anywhere, since even array-api-strict's older standard flags doesn't actually support versions prior to adding complex support. 

We definitely do want to support skipping complex dtypes with ARRAY_API_TESTS_SKIP_DTYPES, but that hasn't been tested before.

cbourjau:
After digging a bit into the hypothesis documentation it seems that using the `nothing` strategy is the way to go here.

cbourjau:
How does this look to you, @asmeurer ?

cbourjau:
@ev-br do you have feedback on this PR?

Issue #312, opened 2024-11-19
test_vecdot does not test for conjugation in the first parameter


asmeurer:
It looks like there isn't any values testing in test_vecdot at all for floating-point arguments. We should add something so that the conjugation can be tested, since this is something many implementations miss (e.g., https://github.com/data-apis/array-api-compat/issues/200, https://github.com/data-apis/array-api-strict/issues/97).

PR #314, opened 2024-11-20
ENH: test vecdot values, incl complex conj

ev-br:
Add a value test for `vecdot` for real and complex fp types: this is a matching test for https://github.com/data-apis/array-api-compat/pull/201 and https://github.com/data-apis/array-api-strict/pull/98

A weird thing is that the test should fail without https://github.com/data-apis/array-api-strict/pull/98 but locally it does not. Probably I'm not generating enough examples---is there a way to bump the number with a cmdline switch? Tried several variations of `python --max-examples=100` but no luck so far.

closes gh-312

### DIFF
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -907,8 +907,8 @@ def test_trace(x, kw):
 
 
 def _conj(x):
-    """Work around xp.conj rejecting floats."""
-    if xp.isdtype(x.dtype, 'complex floating'):
+    # XXX: replace with xp.dtype when all array libraries implement it 
+    if x.dtype in (xp.complex64, xp.complex128):
         return xp.conj(x)
     else:
         return x
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -45,6 +45,7 @@ from . import _array_module
 from . import _array_module as xp
 from ._array_module import linalg
 
+
 def assert_equal(x, y, msg_extra=None):
     extra = '' if not msg_extra else f' ({msg_extra})'
     if x.dtype in dh.all_float_dtypes:
@@ -59,6 +60,7 @@ def assert_equal(x, y, msg_extra=None):
         assert x.dtype == y.dtype, f"The input arrays do not have the same dtype ({x.dtype} != {y.dtype}){extra}"
     else:
         assert_exactly_equal(x, y, msg_extra=msg_extra)
+
 
 def _test_stacks(f, *args, res=None, dims=2, true_val=None,
                  matrix_axes=(-2, -1),
@@ -106,6 +108,7 @@ def _test_stacks(f, *args, res=None, dims=2, true_val=None,
         if true_val:
             assert_equal(decomp_res_stack, true_val(*x_stacks, **kw), msg_extra)
 
+
 def _test_namedtuple(res, fields, func_name):
     """
     Test that res is a namedtuple with the correct fields.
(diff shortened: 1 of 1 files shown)


array_api_tests/test_linalg.py
```diff
@@ -901,6 +905,15 @@ def true_trace(x_stack, offset=0):
 
     _test_stacks(linalg.trace, x, **kw, res=res, dims=0, true_val=true_trace)
 
+
+def _conj(x):
+    """Work around xp.conj rejecting floats."""
+    if xp.isdtype(x.dtype, 'complex floating'):
```
asmeurer, line 911: You won't be able to use this helper here because it won't be implemented in every library. I would suggest adding `is_complex_dtype` to dtype_helpers.py and using it here.

ev-br: Isn't it in the spec? https://data-apis.org/array-api/latest/API_specification/generated/array_api.isdtype.html

asmeurer: If we used `isdtype` here, it would fail with plain PyTorch (for instance), not because `torch.linalg.vecdot` is wrong, but only because PyTorch doesn't implement `isdtype` yet. But that's unnecessary, because we can instead just test against complex dtypes here directly. 

In general in the test suite, we want to make the dependencies across tests as minimal as possible, especially in the case of functions that are definitely not yet implemented everywhere. Tests depending on other array API behavior is unavoidable, but pragmatism is the key point. We know that `dtype in [complex64, complex128]` will work in every library that we test and that `isdtype(dtype, 'complex floating')` will currently not.

ev-br: Ah OK, that's practicality, not purity. That makes total sense, thank you! Made the change.
(somewhat funny how it goes precisely the opposite to some review comments in scipy)

ev-br:
Let's get this in and see if this causes any issues going forward.

array_api_tests/test_linalg.py
asmeurer: Yes, that will probably the hardest thing to get used to, both here in the test suite and also in array-api-strict and array-api-compat. In the test suite, we have to be as minimal as possible with respect to the standard in order to accept as many things as possible. In strict and compat, we have to assume the libraries are what they are (instead of assuming they are array API packages), and wrap them to be more like the array API. In all instances, we consider specific implementation details about specific packages, even when those disagree with the standard.

array_api_tests/test_linalg.py
```diff
@@ -925,11 +938,8 @@ def _test_vecdot(namespace, x1, x2, data):
     ph.assert_result_shape("vecdot", in_shapes=[x1.shape, x2.shape],
                            out_shape=res.shape, expected=expected_shape)
 
-    if x1.dtype in dh.int_dtypes:
-        def true_val(x, y, axis=-1):
-            return xp.sum(xp.multiply(x, y), dtype=res.dtype)
-    else:
-        true_val = None
+    def true_val(x, y, axis=-1):
+        return xp.sum(xp.multiply(_conj(x), y), dtype=res.dtype)
```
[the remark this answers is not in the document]
ev-br: Wait, what: https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/test_linalg.py#L109 and I was sure this PR switches on equality testing. But `assert_equal` does not do what the implies:
https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/test_linalg.py#L49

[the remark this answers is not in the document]
asmeurer: Yes, I originally had it there but later had to comment it out when I found out that you fundamentally can't compare floating point stacks for some functions like eigh because sometimes implementations (like CuPy) use different algorithms that give different (but still mathematically correct) results. https://github.com/data-apis/array-api-tests/pull/101#issuecomment-1111457823 I never renamed the functions. While I did turn the floating-point checks off completely, and they don't make sense for some functions like eigh, they do make sense for functions like vecdot and others where the mathematical answer is has a single well-defined value. We just need to be careful about loss of significance, especially for the tensor contraction functions (vecdot, matmul, tensordot, etc.) that involve additions.

PR #319, opened 2024-11-20
Fix test_reshape

asmeurer:


asmeurer:
This is the failure:

```
______________________________________________________________________________ test_reshape _______________________________________________________________________________

    @pytest.mark.unvectorized
>   @given(
        x=hh.arrays(dtype=hh.all_dtypes, shape=reshape_shape),
        shape=reshape_shapes(reshape_shape),
    )

array_api_tests/test_manipulation_functions.py:366:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ListStrategy(builds(complex, FloatStrategy(min_value=-1.8446742974197924e+19, max_value=1.8446742974197924e+19, allow_...19, allow_nan=False, smallest_nonzero_magnitude=1.1754943508222875e-38).map(downcast)), min_size=8_216, max_size=8_216)
elements = builds(complex, FloatStrategy(min_value=-1.8446742974197924e+19, max_value=1.8446742974197924e+19, allow_nan=False, sm...9, max_value=1.8446742974197924e+19, allow_nan=False, smallest_nonzero_magnitude=1.1754943508222875e-38).map(downcast))
min_size = 8216, max_size = 8216

    def __init__(self, elements, min_size=0, max_size=float("inf")):
        super().__init__()
        self.min_size = min_size or 0
        self.max_size = max_size if max_size is not None else float("inf")
        assert 0 <= self.min_size <= self.max_size
        self.average_size = min(
            max(self.min_size * 2, self.min_size + 5),
            0.5 * (self.min_size + self.max_size),
        )
        self.element_strategy = elements
        if min_size > BUFFER_SIZE:
                f"{self!r} can never generate an example, because min_size is larger "
                "than Hypothesis supports.  Including it is at best slowing down your "
                "tests for no benefit; at worst making them fail (maybe flakily) with "
                "a HealthCheck error."
            )
E           hypothesis.errors.InvalidArgument: ListStrategy(builds(complex, FloatStrategy(min_value=-1.8446742974197924e+19, max_value=1.8446742974197924e+19, allow_nan=False, smallest_nonzero_magnitude=1.1754943508222875e-38).map(downcast), FloatStrategy(min_value=-1.8446742974197924e+19, max_value=1.8446742974197924e+19, allow_nan=False, smallest_nonzero_magnitude=1.1754943508222875e-38).map(downcast)), min_size=8_216, max_size=8_216) can never generate an example, because min_size is larger than Hypothesis supports.  Including it is at best slowing down your tests for no benefit; at worst making them fail (maybe flakily) with a HealthCheck error.
E           while generating 'x' from sampled_from([<class 'numpy.bool_'>, <class 'numpy.uint8'>, <class 'numpy.uint16'>, <class 'numpy.uint32'>, <class 'numpy.uint64'>, <class 'numpy.int8'>, <class 'numpy.int16'>, <class 'numpy.int32'>, <class 'numpy.int64'>, <class 'numpy.float32'>, <class 'numpy.float64'>, <class 'numpy.complex64'>, <class 'numpy.complex128'>]).flatmap(lambda d: arrays(d, *args, elements=elements, **kwargs))

../../miniconda3/envs/array-apis/lib/python3.12/site-packages/hypothesis/strategies/_internal/collections.py:152: InvalidArgument
```

which is the same issue from https://github.com/data-apis/array-api-tests/issues/238. That issue was fixed at https://github.com/data-apis/array-api-tests/pull/242

It's not clear to me how the reshape strategies are generating too large of inputs. However, I do notice some other issues with them, like that the reshape shape is generated using filtering. I don't know if that's related, but it should be fixed anyways.

asmeurer:
I have a fix here. The test should be working properly now. It was generating arrays that were too large. I also went ahead and redid the reshape_shapes strategy to generate reshape compatible shapes more directly. This approach could in principle be applied more broadly wherever we do a filter on size, although filtering on a max size (inequality) isn't as big of a deal as trying to filter against an exact size.

ev-br:
This is great, let's land it straight away.

PR #320, opened 2024-11-21
ENH: add a test for can_cast(complex dtypes)

ev-br:
cross-ref https://github.com/data-apis/array-api-tests/issues/301

### DIFF
--- a/array_api_tests/test_data_type_functions.py
+++ b/array_api_tests/test_data_type_functions.py
@@ -19,16 +19,8 @@ def non_complex_dtypes():
     return xps.boolean_dtypes() | hh.real_dtypes
 
 
-def numeric_dtypes():
-    return xps.boolean_dtypes() | hh.real_dtypes | hh.complex_dtypes
-
-
 def float32(n: Union[int, float]) -> float:
     return struct.unpack("!f", struct.pack("!f", float(n)))[0]
-
-
-def _float_match_complex(complex_dtype):
-    return xp.float32 if complex_dtype == xp.complex64 else xp.float64
 
 
 @given(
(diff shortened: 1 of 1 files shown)


array_api_tests/test_data_type_functions.py
```diff
@@ -19,10 +19,18 @@ def non_complex_dtypes():
     return xps.boolean_dtypes() | hh.real_dtypes
 
 
+def numeric_dtypes():
```
asmeurer, line 22: There's already hh.numeric_dtypes, which is very different from what you're defining here ("numeric dtype" is a term from the standard). 

At any rate, can_cast is supposed to work with all dtypes, so this should be using `hh.all_dtypes`. https://data-apis.org/array-api/latest/API_specification/generated/array_api.can_cast.html#can-cast

array_api_tests/test_data_type_functions.py
```diff
@@ -127,8 +135,15 @@ def test_can_cast(_from, to, data):
...
+            from_dtype = (_float_match_complex(_from)
+                          if _from in (xp.complex64, xp.complex128)
+                          else _from)
+            to_dtype = (_float_match_complex(to)
+                        if to in (xp.complex64, xp.complex128)
+                        else to)
+
+            from_min, from_max = dh.dtype_ranges[from_dtype]
+            to_min, to_max = dh.dtype_ranges[to_dtype]
             expected = from_min >= to_min and from_max <= to_max
```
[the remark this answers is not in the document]
ev-br: Well, I don't know why it was written this way. At a guess, could be related to this wording from https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion

> Behavior is also not specified for integers outside of the bounds of a given integer data type. Integers outside of bounds may result in overflow or an error.

But AFAICS it's only related to python scalars, which we're not testing here anyway. So the last commit removes this logic altogether.

asmeurer:
This looks much simpler now.

array_api_tests/test_data_type_functions.py
```diff
@@ -107,35 +107,20 @@ def test_broadcast_to(x, data):
...
-        if same_family:
-            from_min, from_max = dh.dtype_ranges[_from]
-            to_min, to_max = dh.dtype_ranges[to]
-            expected = from_min >= to_min and from_max <= to_max
-        else:
-            expected = False
     if expected:
         # cross-kind casting is not explicitly disallowed. We can only test
-        # the cases where it should return True. TODO: if expected=False,
-        # check that the array library actually allows such casts.
```
asmeurer, line 138: We could in principle do this (can_cast should really match what casting the functions in the API allow). I don't know if this sort of test is feasible, though.

ev-br: Okay, restored the `TODO` code comment. I've no idea either if this sort of test is feasible or not, so let's kick this can down the road.

ev-br:
Let's get this in, and see if this causes issues down the road. Thank you Aaron for the review.

Issue #284, opened 2024-08-01
Special case tests need to be updated


asmeurer:
Right now there are quite a few warnings printed about failing special case parsing:

```
array_api_tests/test_special_cases.py:711
  /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_special_cases.py:711: UserWarning: case for abs not machine-readable: 'If ``a`` is either ``+infinity`` or ``-infinity`` and ``b`` is any value (including ``NaN``), the result is ``+infinity``'
    warn(f"case for {func_name} not machine-readable: '{case_str}'")

array_api_tests/test_special_cases.py:711
  /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_special_cases.py:711: UserWarning: case for abs not machine-readable: 'If ``a`` is any value (including ``NaN``) and ``b`` is either ``+infinity`` or ``-infinity``, the result is ``+infinity``'
    warn(f"case for {func_name} not machine-readable: '{case_str}'")

array_api_tests/test_special_cases.py:711
  /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_special_cases.py:711: UserWarning: case for abs not machine-readable: 'If ``a`` is either ``+0`` or ``-0``, the result is equal to ``abs(b)``'
    warn(f"case for {func_name} not machine-readable: '{case_str}'")

array_api_tests/test_special_cases.py:711
  /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_special_cases.py:711: UserWarning: case for abs not machine-readable: 'If ``b`` is either ``+0`` or ``-0``, the result is equal to ``abs(a)``'
    warn(f"case for {func_name} not machine-readable: '{case_str}'")

array_api_tests/test_special_cases.py:711
  /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_special_cases.py:711: UserWarning: case for abs not machine-readable: 'If ``a`` is ``NaN`` and ``b`` is a finite number, the result is ``NaN``'
    warn(f"case for {func_name} not machine-readable: '{case_str}'")

array_api_tests/test_special_cases.py:711
  /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/test_special_cases.py:711: UserWarning: case for abs not machine-readable: 'If ``a`` is a finite number and ``b`` is ``NaN``, the result is ``NaN``'
    warn(f"case for {func_name} not machine-readable: '{case_str}'")

[...]

Issue #168, opened 2023-02-03
test_sum can fail with unstable summation


asmeurer:
This test fails with pytorch's sum() (note you need to use 

```
E           AssertionError: out=5.0, but should be roughly 7.0 [sum()]
E           Falsifying example: test_sum(
E               x=tensor([[ 0.0000e+00,  3.3554e+07, -3.3554e+07],
E                       [ 1.0000e+00,  1.0000e+00,  1.0000e+00]]), data=data(...),
E           )
E           Draw 1 (kw): {}
```

Note that you need to use https://github.com/data-apis/array-api-compat/pull/14 because torch.sum has other issues with its signature and type promotion that will fail the test earlier. 

torch truncates the float values, but here is the full tensor:

```py
torch.tensor([[ 0.0000000000e+00,  3.3554432000e+07, -3.3554428000e+07],[ 1.0000000000e+00,  1.0000000000e+00,  1.0000000000e+00]])
```

Note that this is float32, the torch default dtype.  The problem is:

```py
>>> np.float32(33554432.0) + np.float32(1.0)
33554432.0
```

So if you add the elements in the wrong order, you get the wrong sum (5 instead of 7). 

NumPy's sum does this correctly. I don't know if we run the tests on NumPy for long enough if it will also come up with a similar situation. It's really hard to do summation stably. My guess is that the only reason NumPy passes is because NumPy happens to do exactly what we are doing in the test, i.e., upcast to float64 (i.e., `float`), and sum in C-order. But PyTorch's default dtype is float32, so it isn't required to return a float64 from sum(dtype=None) and therefore doesn't upcast internally. 

The spec doesn't require `sum` to use a stable algorithm (like Kahan) vs. naive term-wise summation, and doesn't require the terms to be summed in any particular order. So we should avoid generating examples that can lead to loss of significance. This might require some research, but I think it should be sufficient to avoid generating examples that are too far from each other in magnitude. Another idea would be to compute the exact sum (e.g., using `Fraction`) and compare it to a naive summation. One problem is that the amount of loss of significance depends on the order of summation, and there's no guarantee in which order the library will sum the array.

Also, even if we do this, it's problematic to check the summation by first upcasting to `float` when the input is float32, because that will inherently produce a more accurate result than if the algorithm worked directly on the float32.

lezcano:
[...]

PR #323, opened 2024-11-26
ENH: add crude assert_allclose; use value testing in vecdot

ev-br:


ev-br:
This should fail with array_api_strict until https://github.com/data-apis/array-api-strict/pull/98 lands.
(verified locally that it passes with numpy and fails with array-api-strict)

asmeurer:
The CI here passes. It might be because the CI doesn't run enough examples by default and didn't catch the error.

asmeurer:
It would be a good idea to run this with a lot of examples. My worry is that we will see the same sorts of flaky falure issues we saw with https://github.com/data-apis/array-api-tests/issues/168 due to loss of significance unless we filter out ill conditioned inputs (as in https://github.com/data-apis/array-api-tests/pull/290). 

Of course, we can always just merge this and revert it or fix it if issues crop up. The CI here doesn't run many examples relatively speaking, but the CI for array-api-strict and array-api-compat is where you usually find the errors, because they have such a large matrix that they rarer errors are very likely to show up (that is assuming the test in question isn't already being XFAILed for some reason).

array_api_tests/pytest_helpers.py
```diff
@@ -599,3 +600,18 @@ def assert_array_elements(
             at_expected = expected[idx]
             msg = msg_template.format(sh.fmt_idx(out_repr, idx), at_out, at_expected)
             assert at_out == at_expected, msg
+
+
+def assert_allclose(actual, desired, *, atol=1e-7, rtol=1e-7, equal_nan=True, msg_extra=""):
```
asmeurer, line 605: There's also some similar helpers in test_operators_and_elementwise_functions.py (that file has some of the more complex helper functions, since most of the tests are very similar).

ev-br:
An alternative strategy, in view of https://github.com/data-apis/array-api-tests/pull/314#discussion_r1859369692, could be to only test "exact" things in hypothesis (rename `assert_equal` to `assert_equal_if_int` or some such) and add fp value testing only for specific functions with fixed inputs, as regular regression tests without hypothesis.

asmeurer:
Yes, it's a possibility. It's not something we've really done yet, but maybe we should think about it (see also https://github.com/data-apis/array-api-tests/issues/284). Non-hypothesis tests have their own issues, which is that they only test specific things. The hypothesis tests have been pretty good at checking a wide range of cases which would definitely be missed with traditional testing. I would definitely strongly encourage the majority of the test suite to continue to use hypothesis. Even something simple like a function with `n` keyword arguments has something like $2^n$ possible combinations of inputs, and in many cases these combinations do materially affect each other. 

And I do agree we should rename functions to be more descriptive of what they are actually doing.

PR #324, opened 2024-11-29
Add testing of `reciprocal`, `cumulative_prod` from 2024.12 spec revision

ev-br:


ev-br:
The CI failure is #332. More importantly though, the new tests did not run because they are decorated for 2024.12 spec revision.

ev-br:
Verified that the new tests are running now that 2024.12 exists and `array-api-strict==2.3` is up on PyPI. Merging.

PR #332, opened 2024-12-28
CI: update python versions for lint.yml

ev-br:
Python 3.8 is not available on ubuntu-latest images anymore.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,15 +6,11 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.10", "3.11"]
-
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


ev-br:
I've no idea where the stalling "buld: (3.8)" and "build: (3.9)" jobs are coming from. Clearing the actions cache does not help. Unless it's going to somehow resolve itself, the "lint" CI job in in this repo is broken.  Maybe not that big of a loss anywhay.

rgommers:
Not sure it's the root cause, but `setup-python@v1` is very old. Maybe update actions to latest version?

rgommers:
Linting on two separate Python versions isn't useful I think, so also better to remove the matrix-ing over Python versions while you're at it.

rgommers:
And Ubuntu-provided packages shouldn't matter? The Python comes from `setup-python`, which supports 3.8/3.9 just fine.

ev-br:
Good idea, thanks. Updated the action versions, removed the matrix. Stalling builds are still there.

rgommers:
I can't find the problem so quickly. Let's see if it resolves itself now that the job is updated.

PR #335, opened 2025-01-16
TST: Test copy behavior in astype

lithomas1:


array_api_tests/pytest_helpers.py
```diff
@@ -483,6 +487,48 @@ def assert_fill(
...
+        return cmath.isnan(s2)
+    else:
+        return s1 == s2
+
+
+def assert_kw_copy(func_name, x, out, data, copy):
+    """
+    Assert copy=True/False functionality is respected
+
+    TODO: we're not able to check scalars with this approach
```
crusaderky, line 501: We should definitely be able to check 0-dimensional arrays. What we can't check is np.generic, whiich AFAIK is a numpy-specific type

crusaderky, line 501: This doesn't look right.
- `astype(copy=True)` should behave like `asarray(copy=True)`.
- `astype(copy=False)` should behave like `asarray(copy=None)`.
- `asarray(copy=False)` has no matching behaviour in `astype`.

crusaderky (changes requested):

crusaderky:
@lithomas1 any update on this?

lithomas1:
> @lithomas1 any update on this?

Planning on circling back since the dask PR is merged now.

PR #334, opened 2024-12-30
feat: `ARRAY_API_TESTS_MODULE` for runtime-defined xp

lucascolley:

PR #336, opened 2025-01-24
feat: `ARRAY_API_TESTS_MODULE` for runtime-defined xp

lucascolley:
take 2 of gh-334 now that CI is fixed


lucascolley:
@rgommers it looks like the <py310 jobs are still expected

PR #340, opened 2025-02-02
ENH: default to 2024.12 revision

ev-br:
TODO:

- [x] wait until 2024.12 revision is actually released :-).
- [x] coordinate with https://github.com/HypothesisWorks/hypothesis/pull/4262


ev-br:
The hypothesis update is up on PyPI as v6.125.0 (wow), the CI failure now is the lack of `2024.12` revision itsefl in the spec.

PR #342, opened 2025-02-18
ENH: test vecdot conjugation

ev-br:
A small no-hypothsis test to make sure vecdot conjugates its first argument

This is a light-touch alternative to https://github.com/data-apis/array-api-tests/pull/323, which is blocked by the difficulties of value testing with hypothesis.

TODO:

- [ ] update array-api-strict to `>= 2.3.0`, when available.

### DIFF
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -964,6 +964,16 @@ def test_vecdot(x1, x2, data):
     _test_vecdot(_array_module, x1, x2, data)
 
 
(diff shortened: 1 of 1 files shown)


ev-br:
CI fails `with array-api-strict==2.2`, passes locally with `array-api-strict==2.3dev` since the fix in `-strict` landed after 2.2 was cut.

ev-br:
Now that `array-api-strict==2.3` is released, tests pass. Merging.

PR #344, opened 2025-02-26
Add regex support to `--xfails-file`

crusaderky:
As discussed in https://github.com/data-apis/array-api-compat/pull/253
@ev-br

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import argparse
 import warnings
 import os
+import re
 
 from hypothesis import settings
 from hypothesis.errors import InvalidArgument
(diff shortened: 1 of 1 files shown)


ev-br:
Before trying to understand how it works, I'd like to understand how it is meant to help data-apis/array-api-compat#253. Mind elaborating please @crusaderky ?

crusaderky:
The original intent was to add to the xfails file
```
.*not implemented for 'UInt16'
```
However I now realise that
- this PR wouldn't achieve that: the match is on test name+parametrize; not on error raised
- the dtype iteration is performed by hypothesis and not by parametrize which means ignoring such an error would neuter the test for all _other_ dtypes.

PR #349, opened 2025-03-16
ENH: improve testing of `result_type`

ev-br:
cross-ref https://github.com/scipy/scipy/pull/22695

- [x] add a test that result_type does not depend on the order of arguments
- [x] add a test with arrays and dtypes
- [x] add a test with arrays, dtypes and scalars
- [ ] <s>check if we can test with f64 being the pytorch default dtype</s>: tracked in https://github.com/data-apis/array-api-compat/issues/278


ev-br:
This adds testing motivated by https://github.com/data-apis/array-api-compat/issues/274 and https://github.com/data-apis/array-api-compat/issues/273. Would be great if you could test it @mdhaber

PR #350, opened 2025-03-17
DOC: add quotes around a shell command

ev-br:
otherwise, it barfs with `bash: syntax error near unexpected token `('`


lucascolley (approved):
thanks!

PR #351, opened 2025-03-20
BUG: work around some names not available in 2022.12

ev-br:
Case in point: `xp.hypot` is only available in 2023.12, and has gotten the support for scalars in 2024.12.

The test for scalars is skipped for versions older than 2024.12, but the test parametrization is evaluated _before_ the skip. Therefore we cannot just use `xp.hypot` in `@parametrize`.


ev-br:
Merging straight away to unblock https://github.com/data-apis/array-api-strict/pull/133

Issue #339, opened 2025-01-30
Failing `np.clip()` on float32


crusaderky:
e.g. https://github.com/data-apis/array-api-compat/actions/runs/13060007276/job/36440413878?pr=247

```
FAILED array_api_tests/test_operators_and_elementwise_functions.py::test_clip - AssertionError: out[()]=-16777216.0 but should be -16777217.0 [clip()]
  x[()]=0.0, min[()]=None, max[()]=-16777217.0
assert -16777216.0 == np.float64(-16777217.0)
Falsifying example: test_clip(
    x=array(0., dtype=float32),
    data=data(...),
)
Draw 1 (min.shape, max.shape): ((), ())
Draw 2 (min): None
Draw 3 (max): np.float64(-16777217.0)
Draw 4 (kwargs): {'max': np.float64(-16777217.0)}
```

The problem is somehow caused by the fact that `np.float32(-16777217.0) == -16777216.0`.
However I can't reproduce the error on numpy 2.0.2:

```python
>>> np.clip(np.asarray(0, dtype=np.float32), None, -16777217.0)
np.float32(-16777216.0)  # Correctly follows rule about binops mixing pure python and arrays

>>> np.clip(np.asarray(0, dtype=np.float32), None, np.float32(-16777217.0))
np.float32(-16777216.0)

>>> np.clip(np.asarray(0, dtype=np.float32), None, np.float64(-16777217.0))
np.float64(-16777217.0)
```

ev-br:
managed to catch it on numpy 2.2. A fix in gh-355

PR #355, opened 2025-03-23
BUG: fix flaky clip with float32

ev-br:
Testing `clip` converts array elements to builtin types and uses exact equaity. Converting `out` from float32 to 64-bit builtin float incurs floating-point errors, and further exact equality testing fails. So use `math.isclose` instead.

EDIT: does not close gh-339

### DIFF
--- a/array_api_tests/test_operators_and_elementwise_functions.py
+++ b/array_api_tests/test_operators_and_elementwise_functions.py
@@ -26,6 +26,9 @@ from .typing import Array, DataType, Param, Scalar, ScalarType, Shape
 pytestmark = pytest.mark.unvectorized
 
 
+EPS32 = xp.finfo(xp.float32).eps
+
+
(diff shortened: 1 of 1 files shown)


crusaderky:
This doesn't look right.

To my understanding, the bug report implies that np.clip(x: float32, min: float64, max: float64) is downgrading min/max to float32, then doing the arithmetics in float32, and finally upgrading everything back to float64, which is definitely wrong.

This PR shouldn't do anything, because in the failure out.dtype=float64.

ev-br:
The failure I caught was exactly with out.dtype == float32

crusaderky:
Then this won't close #339, as there the output must have had output dtype=float64: ```
Draw 4 (kwargs): {'max': np.float64(-16777217.0)}
```

ev-br:
OK, let's keep gh-339 open until somebody can repro

Issue #359, opened 2025-04-08
Issue with test_clip


ndgrigorian:
```
    | Falsifying example: test_clip(
    |     x=array(0, dtype=uint8),
    |     data=data(...),
    | )
    | Draw 1 (min.shape, max.shape): ((), ())
    | Draw 2 (min): array(0, dtype=uint8)
    | Draw 3 (max): array(0, dtype=uint16)
    | Draw 4 (kwargs): {'min': array(0, dtype=uint8), 'max': array(0, dtype=uint16)}
```

Per spec,
> If either min or max is an array having a different data type than x, behavior is unspecified and thus implementation-dependent.

> Clarified that behavior is only defined when x, min, and max resolve to arrays having the same data type.

It seems to me that this is a malformed test, and only cases where `min` and `max` are (or resolve to, but even that's a stretch per first condition) the type of `x` should be tested.

ev-br:
Agreed, the 2024 spec clarification is pretty clear on arrays of different dtypes. There's also gh-339 which report issues for mixing float32 and float64. Here's a fix: gh-360

PR #360, opened 2025-04-08
test_clip: make sure `min` and `max` have the same dtype as `x`

ev-br:
closes gh-359, closes gh-339

Per the spec: 'should have the same dtype as x': https://data-apis.org/array-api/latest/API_specification/generated/array_api.clip.html

The spec is not very clear on whether `min` and `max` can be `int` scalars for a real-valued `x` array, should probably clarify in the spec:

- On one hand, _Hence, if x and either min or max have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent._
- On the other hand, https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion, allows int->float->float32

### DIFF
--- a/array_api_tests/test_operators_and_elementwise_functions.py
+++ b/array_api_tests/test_operators_and_elementwise_functions.py
@@ -987,17 +987,15 @@ def test_clip(x, data):
                                                                 base_shape=x.shape),
                                 label="min.shape, max.shape")
 
-    dtypes = hh.real_floating_dtypes if dh.is_float_dtype(x.dtype) else hh.int_dtypes
-
     min = data.draw(st.one_of(
         st.none(),
         hh.scalars(dtypes=st.just(x.dtype)),
-        hh.arrays(dtype=dtypes, shape=shape1),
+        hh.arrays(dtype=st.just(x.dtype), shape=shape1),
     ), label="min")
     max = data.draw(st.one_of(
         st.none(),
         hh.scalars(dtypes=st.just(x.dtype)),
-        hh.arrays(dtype=dtypes, shape=shape2),
+        hh.arrays(dtype=st.just(x.dtype), shape=shape2),
     ), label="max")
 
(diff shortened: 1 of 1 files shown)


ev-br:
Run tests with `array_api_compat.{numpy,torch}` and `jax.numpy` with 50_000 examples locally.

oleksandr-pavlyk (approved):

crusaderky:
> The spec is not very clear on whether `min` and `max` can be `int` scalars for a real-valued `x` array, should probably clarify in the spec:
> 
> * On one hand, _Hence, if x and either min or max have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent._
> * On the other hand, https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion, allows int->float->float32

Definitely the spec should clarify that int scalars should be cast to float.

crusaderky:
https://github.com/data-apis/array-api/issues/925

array_api_tests/test_operators_and_elementwise_functions.py
```diff
@@ -987,17 +987,15 @@ def test_clip(x, data):
                                                                 base_shape=x.shape),
                                 label="min.shape, max.shape")
 
-    dtypes = hh.real_floating_dtypes if dh.is_float_dtype(x.dtype) else hh.int_dtypes
-
     min = data.draw(st.one_of(
         st.none(),
         hh.scalars(dtypes=st.just(x.dtype)),
```
crusaderky, line 992: note: this misses out on the use case `clip(Array[float], int)`

ev-br:
Okay, let's merge this as approved for this seems to fix rather annoying failures. 
I've added a bullet point in https://github.com/data-apis/array-api-tests/issues/301 to track the "int min, float array" case.
Thanks @ndgrigorian @crusaderky  for the reviews.

PR #362, opened 2025-04-09
ENH: some more tests of inspection functions

ev-br:


ev-br:
This runs into https://github.com/data-apis/array-api/issues/923 on `jax.numpy`.

Issue #299, opened 2024-11-14
Add tests for new 2024.12 functionality


asmeurer:
This is a tracking issue for testing new 2024.12 functionality. 

There isn't a changelog yet, but the new functionality that I'm aware of is:

**New functions**

- [x] `diff` (https://github.com/data-apis/array-api-tests/pull/369)
- [x] `nextafter` (https://github.com/data-apis/array-api-tests/pull/366)
- [x] `reciprocal`
- [x] `take_along_axis` (https://github.com/data-apis/array-api-tests/pull/367)
- [x] `count_nonzero`
- [x] `cumulative_prod`

**New Functionality**

- [x] The `'max dimensions'` key of `__array_namespace_info__().capabilities()` (note that `capabilities` itself is not yet tested at all, see https://github.com/data-apis/array-api-tests/issues/249) -- some smoke-testing in gh-362

**Scalar support:** most functions are (lightly) tested. After gh-348, the following binary functions are still not tested at all:

- [x] `bitwise_right_shift`  (https://github.com/data-apis/array-api-tests/pull/371)
- [x] `bitwise_left_shift`  (https://github.com/data-apis/array-api-tests/pull/371)
 - [x] `nextafter` (https://github.com/data-apis/array-api-tests/pull/366)
 - [x] `where` (https://github.com/data-apis/array-api-tests/pull/370)
 - [x] `pow` (https://github.com/data-apis/array-api-tests/pull/371)
 - [x] `result_type` ( gh-349 )

ev-br:
cross-ref https://github.com/data-apis/array-api-tests/pull/324 for testing `reciprocal`.

ev-br:
Additionally, 

- [ ] `count_nonzero`
- [ ] `cumulative_prod`

ev-br:
https://github.com/data-apis/array-api-tests/pull/347 adds a test for `count_nonzero`

ev-br:
We now officially have the 2024.12 test coverage. At least all new functions are covered, with remaining TODOs tracked in gh-301 and gh-364.

PR #367, opened 2025-04-21
ENH: test `take_along_axis`

ev-br:
Towards gh-299: add testing of `take_along_axis`.

### DIFF
--- a/array_api_tests/test_indexing_functions.py
+++ b/array_api_tests/test_indexing_functions.py
@@ -73,22 +73,30 @@ def test_take_along_axis(x, data):
     # TODO
     # 2. negative indices
     # 3. different dtypes for indices
-    axis = data.draw(st.integers(-x.ndim, max(x.ndim - 1, 0)), label="axis")
-    len_axis = data.draw(st.integers(0, 2*x.shape[axis]), label="len_axis")
+    axis = data.draw(
+        st.integers(-x.ndim, max(x.ndim - 1, 0)) | st.none(),
+        label="axis"
+    )
+    if axis is None:
+        axis_kw = {}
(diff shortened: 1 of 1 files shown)


ev-br:
The spec requires that `axis=-1` has a default value. Pretty much all libraries differ: 

- `jax.numpy` : `axis` is a required argument. jax does not support the 2024.12 revision yet though.
- `numpy`: `axis` is a required argument.
- `cupy`: `axis` is a required argument
- `torch`: `dim=None` defaults to None, but the documentation states that `dim=None` ravels.
- `dask.array`: does not implement `take_along_axis` at all.

https://github.com/data-apis/array-api-compat/pull/317 works around it for numpy and cupy; torch was alread covered; dask.array will need a test skip in https://github.com/data-apis/array-api-compat/pull/317.

cc @crusaderky  for dask

ev-br:
Ran locally the combination of this test and https://github.com/data-apis/array-api-compat/pull/317, for 10_000 examples on each of `torch`, `cupy` and `numpy`.

```
$ ARRAY_API_TESTS_SKIP_DTYPES=uint16,uint32,uint64 ARRAY_API_TESTS_MODULE=array_api_compat.torch pytest array_api_tests/test_indexing_functions.py::test_take_along_axis -v  --max-examples=10_000
```

crusaderky:
Don't expect a dask implementation anytime soon: https://github.com/dask/dask/issues/3663
Best we can do is hack together something in array-api-compat that rechunks to a single chunk, much like it already happens for `sort` and `argsort`. This will work on a "better than nothing" basis, with very big performance caveats.

ev-br:
Thanks, will add a skip with the link to the dask issue then.

PR #371, opened 2025-05-05
test pow() with scalars

ev-br:
towards https://github.com/data-apis/array-api-tests/issues/299


ev-br:
Ran 20_000 examples locally with jax.numpy, cupy, torch, numpy and array_api_strict.
The only issue is that
`test_binary_with_scalars_real[pow]` needs to be xfailed under CuPy 13.x, which does not fully implement NEP 50. The issue with `pow` is fixed in CuPy 14.0.0a1, so will be able to remove the xfail when 14.x becomes the minimum supported CuPy version. Until then, it's an xfail; which I'll add in a separate PR to array-api-compat.

PR #373, opened 2025-05-11
ENH: add ARRAY_API_TESTS_XFAIL_MARK to turn xfails into skips

ev-br:
Locally, this speeds the torch test run by a factor of 4 or more.


ev-br:
Turning xfails into skips looks like a no-brainer, based on https://github.com/data-apis/array-api-compat/pull/321

Am now checking the effect of lowering `MAX_ARRAY_SIZE` down to 1024 from 10_000. This is expected to help with elementwise array iterations.

PR #376, opened 2025-05-14
ENH: test_tile: check the shapes

ev-br:
towards gh-301


ev-br:
A minor addition, merging.

Issue #379, opened 2025-05-21
Providing reproduction code for failed test cases


mtsokol:
Hi all!

Here's one idea on how to potentially improve interacting with failing tests in `array-api-tests` suite which @ev-br and I discussed this week.

---

I've been involved in the introduction of `array-api-tests` suite in a few repositories and, from my personal experience, the activity that took the most of the time was figuring out root causes of failed tests.

Each failure shows the stack trace (or multiple stack traces, if there were e.g. 3 distinct failures for a given test), but at times they are merely related to the actual root cause. For instance, in NumPy we have an `xfail`:
```
# fails on np.repeat(np.array([]), np.array([])) edge test case
array_api_tests/test_manipulation_functions.py::test_repeat
```
which originally was reported by the test suite with an error:
```
Cannot cast array data from dtype('float64') to dtype('int64')
```
I had to manually recreate the function call with the exact inputs to understand which edge case we hit. And for each test the error message was either accurate, like missing keyword argument, or irrelevant - which was mostly related to array scalars, Python builtins, or 0-D arrays as inputs or outputs. 

From my point of view, one possible improvement of this process could be something like a CLI option `--with-repro-snippets`, where each failing test is accompanied by a copy-paste line that exactly reproduces it.

So for `test_repeat` I would get:
```
Here's a line that reproduces it:
xp.repeat(xp.asarray([]), xp.asarray([]))
```
The `array-api-tests` suite would compose it with `f"xp.{func_name}({inputs}, {kwargs})"` when a function call fails.

This way, when I run a test suite for the first time and get e.g. 50 failures, I can iterate a bit faster by skipping the "reproduce failing code snippet" step. WDYT? Please share your thoughts!

---

More of a nitpick but some Array API functions are used in multiple tests, like `reshape`. I think it’s used primarily for setting up inputs for tests. When that one function is missing from the namespace, a large part of the test suite fails with more cryptic error messages. I think that there could be a `hasattr(xp, "reshape")` decorator for tests and if a function for setting inputs is missing, the error says so.

mtsokol:
> the current `ph.assert_*` helpers go to incredible lengths to give useful diagnostics. Maybe we could improve specific helpers? You mention `test_repeat`, do you happen to remember other cases with less-than-useful diagnostics?

I can take a look at the tests that I've fixed in the past and provide a few more!

[...]

Issue #51, opened 2021-12-16
Specify dependency of each test case


honno:
The test suite is mostly made up of test methods for each function (or array object method) in the spec. The majority of these tests require other functions to do everything we want, which is problematic when an underlying function does not work as you'll get a lot of spammy that actually relate to a single fundamental problem (or even false positives).

So I think it would be a good idea if _we declared the dependencies of each test method_, i.e. the functions we use and assume has correct behaviour. We could use these declarations to create a **dependency graph**, which could be hooked with pytest to prioritise zero/low-dependency tests first, and by default skipping tests that use functions we've deemed incorrect. This will benefit:

1. Array API adoptors, who would much more easily see what they should prioritise developing
2. Us, who'd be able to see areas where we can try to cut down the functions we depend on

Ideas for declaring dependencies per test method:

- Decorator which lists function and method names
- Automatically infer dependencies via looking at function code (either searching the code as string, or using the AST)

honno:
I still think this would be nice, but I think the test suite could do with some more refinement generally for this to be feasible, and even then doesn't seem to be as big a deal as I made out it out to be. Every time I've seen the test suite used, devs are usually identifying that buggy fundamental components are preventing testing other things... it's just that it probably took them way longer than ideal heh.

Zac-HD:
The 80/20 of this would be to group tests into `core`, `standard`, and then `extension-xxx` groups - I think this would be fairly easy with a module-level `pytestmark = pytest.mark.core` (etc).  Then if you run into a problem, try `pytest -m core` and then `pytest -m "core or standard"` to check whether it's _something_ foundational.

Hypothesis takes a similar approach using directories to organize our selftests (core=conjecture, cover + nocover, extensions, etc.) which works pretty well at higher volumes.  Exact dependency tracking is a lovely idea in principle but AFAICT very rare in practice because it's not usually worth the trouble 😅

PR #380, opened 2025-05-26
MAINT: Check essential data functions

mtsokol:
Hi @ev-br,

As discussed in #379, here's a PR that introduces `has_setup_funcs` mark.

I applied it to the first batch of tests that use `.draw()` calls, which underneath uses `do_draw`, so at least one of the `asarray`, `isnan`, `reshape`, `zeros` is used (which `has_setup_funcs` checks).

I still need to find all tests that use `draw`.

### DIFF
--- a/array_api_tests/test_creation_functions.py
+++ b/array_api_tests/test_creation_functions.py
@@ -3,6 +3,7 @@ import math
 from itertools import count
 from typing import Iterator, NamedTuple, Union
 
+import pytest
 from hypothesis import assume, given, note
 from hypothesis import strategies as st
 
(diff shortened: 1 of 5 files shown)


ev-br:
This is a great start!

I think it's best to generalize it from the onset to address https://github.com/data-apis/array-api-tests/issues/51
Ideally, there's a decorator (or a specific pytest mark, whichever is easier) with arguments. Tentatively calling this decorator `def require(*args: str)`, 

```
@require('sin', 'zeros_like')
@given(....)
def test_func(...):
   ....
``` 

so this would skip the test unless `xp.sin` and `xp.zeros_like` are available. 

Whether the list dependencies of  `draw` (your `essential_functions`) should be automatically added or needs listed explicitly --- so that the above example is `@require(draw_deps, 'sin', 'zeros_like')` --- this can go both ways. I'd lean towards explicit is better than implicit, but it's a mild preference.

Also do you think it'd be useful to add a knob to turn the check off, say, with env variable? The reason is if a library does not implement anything at all, is the test suite all green (everything is skipped) or all red (nothing works).

ev-br:
From working on MLX compatibility, two other essential functions are `__complex__` (otherwise something fails deep inside hypothesis), and the inspection helper, `__array_namespace_info__` (this one is needed for dlpack testing, and is imported at the module level, so the test suite does not run at all if missing.

Issue #300, opened 2024-11-14
Skipped flaky tests


asmeurer:
Several tests are completely skipped right now because they are "flaky". 

- `test_reshape`
- `test_std`
- `test_var`
- `test_remainder`

This is a pretty high priority issue because these functions are effectively completely untested, even though they appear to be tested. 

Tests should be written in such a way that they aren't flaky, for instance, by using high numerical tolerances (or if necessary, avoiding values testing entirely). 

Note that health checks for timeouts should just be skipped, and health checks for filtering too much should be fixed by fixing the strategy.

ev-br:
Looking at `test_std`, https://github.com/data-apis/array-api-tests/blob/master/array_api_tests/test_statistical_functions.py#L262, it does not seem to attempt any value testing. Then what is flaky, `assert_dtype` or `assert_keepdimable_shape`?

asmeurer:
I've no idea what's flaky with any of these. The first order of business would to remove that decorator and figure out why the test was failing. It's also possible that some of these were only flaky with certain libraries.

asmeurer:
Also, it's possible the flakyness was fixed and the skip was never removed. It looks like skip was added in https://github.com/data-apis/array-api-tests/pull/233 (with no explanation) if you want to check previous versions. 

At best, if the test seems to be passing, we can just remove the skip and see if any upstream failures are found. Like I mentioned in another issue, it's really easy to just revert changes here if they break stuff since we don't even have releases, so I wouldn't be *too* worried about that.

ev-br:
`test_reshape` is fixed in gh-319

ev-br:
Caught a `test_std` failure with `array_api_compat.numpy`:


```
array_api_tests/test_statistical_functions.py::test_std FAILED                   [100%]

    @given(
>       x=hh.arrays(
            dtype=hh.real_floating_dtypes,
            shape=hh.shapes(min_side=1),
            elements={"allow_nan": False},
        ).filter(lambda x: math.prod(x.shape) >= 2),
        data=st.data(),
    )

array_api_tests/test_statistical_functions.py:254: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def inconsistent_generation():
>       raise FlakyStrategyDefinition(
            "Inconsistent data generation! Data generation behaved differently "
[...]

PR #384, opened 2025-06-06
POC: start adding "repro snippets" for failed assertions

ev-br:
It is not always easy to tell what exactly failed with hypothesis (gh-379). Thus add a failing snippet to the exceptions' output.

With this PR, the OP example from gh-379 preserves whatever output pytest and hypothesis generated, and adds an explicit incantation that triggered the error:

```
$ ARRAY_API_TESTS_MODULE=array_api_compat.numpy pytest array_api_tests/test_manipulation_functions.py::test_repeat

... snip ...

array_api_tests/test_manipulation_functions.py:288: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
array_api_tests/test_manipulation_functions.py:318: in test_repeat
    out = xp.repeat(x, repeats, **kw)
../../miniforge3/envs/array-api/lib/python3.12/site-packages/numpy/_core/fromnumeric.py:506: in repeat
    return _wrapfunc(a, 'repeat', repeats, axis=axis)
../../miniforge3/envs/array-api/lib/python3.12/site-packages/numpy/_core/fromnumeric.py:66: in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = array([], dtype=bool), method = 'repeat', args = (array([], dtype=uint64),)
kwds = {'axis': None}
conv = <numpy._core._multiarray_umath._array_converter object at 0x7fd5f2252c10>
arr = array([], dtype=bool)

    def _wrapit(obj, method, *args, **kwds):
        conv = _array_converter(obj)
        # As this already tried the method, subok is maybe quite reasonable here
        # but this follows what was done before. TODO: revisit this.
        arr, = conv.as_arrays(subok=False)
>       result = getattr(arr, method)(*args, **kwds)
E       TypeError: Cannot cast array data from dtype('uint64') to dtype('int64') according to the rule 'safe'
E       
E       ========== FAILING CODE SNIPPET:
E       xp.repeat(array([], dtype=bool),array([], dtype=uint64), **kw) with kw={}
E       ====================
E       
E       Falsifying example: test_repeat(
E           x=array([], dtype=bool),
E           kw={},
E           data=data(...),
E       )
E       Draw 1 (repeats): array([], dtype=uint64)

../../miniforge3/envs/array-api/lib/python3.12/site-packages/numpy/_core/fromnumeric.py:46: TypeError
=============================== short test summary info ================================
FAILED array_api_tests/test_manipulation_functions.py::test_repeat - TypeError: Cannot cast array data from dtype('uint64') to dtype('int64') according ...
================================== 1 failed in 0.24s ===================================
```

The approach here is a bit manual, and I don't see a generic way to make it less so.

### DIFF
--- a/array_api_tests/pytest_helpers.py
+++ b/array_api_tests/pytest_helpers.py
@@ -599,3 +599,7 @@ def assert_array_elements(
             at_expected = expected[idx]
             msg = msg_template.format(sh.fmt_idx(out_repr, idx), at_out, at_expected)
             assert at_out == at_expected, msg
+
+
+def format_snippet(s: str):
+    return f"\n{'='*10} FAILING CODE SNIPPET:\n{s}\n{'='*20}\n"
--- a/array_api_tests/test_data_type_functions.py
+++ b/array_api_tests/test_data_type_functions.py
@@ -160,13 +160,17 @@ def test_finfo(dtype):
         # np.float64 and np.asarray(1, dtype=np.float64).dtype are different
         xp.asarray(1, dtype=dtype).dtype,
     ):
-        out = xp.finfo(arg)
-        assert isinstance(out.bits, int)
-        assert isinstance(out.eps, float)
-        assert isinstance(out.max, float)
-        assert isinstance(out.min, float)
-        assert isinstance(out.smallest_normal, float)
-
+        repro_snippet = ph.format_snippet(f"xp.finfo({arg})")
+        try:
+            out = xp.finfo(arg)
+            assert isinstance(out.bits, int)
+            assert isinstance(out.eps, float)
+            assert isinstance(out.max, float)
+            assert isinstance(out.min, float)
+            assert isinstance(out.smallest_normal, float)
+        except Exception as exc:
+            exc.add_note(repro_snippet)
+            raise
 
 @pytest.mark.min_version("2022.12")
 @pytest.mark.parametrize("dtype", dh.real_float_dtypes + dh.complex_dtypes)
--- a/array_api_tests/test_manipulation_functions.py
+++ b/array_api_tests/test_manipulation_functions.py
@@ -313,37 +313,45 @@ def test_repeat(x, kw, data):
 
     assume(n_repititions <= hh.SQRT_MAX_ARRAY_SIZE)
 
-    out = xp.repeat(x, repeats, **kw)
-    ph.assert_dtype("repeat", in_dtype=x.dtype, out_dtype=out.dtype)
-    if axis is None:
-        expected_shape = (n_repititions,)
-    else:
-        expected_shape = list(shape)
-        expected_shape[axis] = n_repititions
-        expected_shape = tuple(expected_shape)
-    ph.assert_shape("repeat", out_shape=out.shape, expected=expected_shape)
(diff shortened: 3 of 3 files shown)


ev-br:
- [ ] test_array_object.py
- [ ] test_constants.py
- [ ] test_creation_functions.py
- [ ] test_data_type_functions.py
- [ ] test_fft.py
- [ ] test_has_names.py
- [ ] test_indexing_functions.py
- [ ] test_inspection_functions.py
- [ ] test_linalg.py
- [x] test_manipulation_functions.py
- [ ] test_operators_and_elementwise_functions.py
- [ ] test_searching_functions.py
- [ ] test_set_functions.py
- [ ] test_signatures.py
- [ ] test_sorting_functions.py
- [ ] test_special_cases.py
- [ ] test_statistical_functions.py
- [ ] test_utility_functions.py

ev-br:
The shape of this POC is getting clear now. I've converted one of the test modules, `test_manipulation_functions.py`. The patch size is mostly whitespace/indentation, and is best viewed with the "hide whitespace" option.

Converting the rest is a bit of mechanical work, but is totally doable.

Downsides:
- `git blame` gets a bit messier, will need `git blame -M -C -C ` or some such to help git see past trivial indentation changes.
- this will only catch failures in the functions under tests and test assertions. Tricky failures inside `hypothesis` (either "essential functions" of https://github.com/data-apis/array-api-tests/issues/379#issuecomment-2903993935 or obscure "flaky" failures of https://github.com/data-apis/array-api-tests/issues/300) will not be caught.

ev-br:
Thanks for the ping @cbourjau !
- Which function(s) did you hit this with?
- Is the current use of `repr` appropriate for you? (it makes a difference in numpy, to tell numpy scalars apart from python scalars--- is the `f"{arg!r}"` good enough for ndonnx?)

-----

I suppose the only thing blocking this is some elbow grease, would you be able to help? Would be great to split the set of test files, https://github.com/data-apis/array-api-tests/pull/384#issuecomment-2953843653 --- the changes are pretty mechanical, just need to be test by test source file by source file.

The solution to the git blame issue is to add a `.git-blame-ignore-revs` file to the repo root: 
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view, here's the SciPy usage: https://github.com/scipy/scipy/blob/main/.git-blame-ignore-revs

Therefore,

- it's best to work source file by source file
- to keep the number of entries under control, we should either use a commit per source file or a PR per source file + squash-merge.
- we add the commit hashes to `.git-blame-ignore-revs`

ev-br:
This nearly slipped under the radar. Thanks @cbourjau for the nudge, thanks @mtsokol for the idea. I'll open a tracker issue for test modules to convert.

PR #386, opened 2025-06-17
CI: bump python versions to match the new array-api-strict lower bound

ev-br:


ev-br:
Needed for `array_api_strict==2.4` which bumped the lower bound to python 3.12, thus merging straight away.

PR #388, opened 2025-08-12
Pass `tuple[int]` instead of `tuple[Array]` as shape to `xp.full`

cbourjau:
The existing test for `repeat` relies on `full` being able to consume a tupel of scalar `Array` objects. The latter is not part of the [standard](https://data-apis.org/array-api/draft/API_specification/generated/array_api.full.html#full). A simple call to Python's `int` saves the day.


ev-br (approved):
Thank you @cbourjau !

Issue #389, opened 2025-10-02
argsort: default value of stable not tested


cakedev0:
See issue: https://github.com/data-apis/array-api-compat/issues/354

This section of the [test](https://github.com/data-apis/array-api-tests/blob/1fb61d5593c544a1dda625e307ef6821bcbd905f/array_api_tests/test_sorting_functions.py#L41):
```Python
    kw = data.draw(
        hh.kwargs(
            axis=st.integers(-x.ndim, x.ndim - 1),
            descending=st.booleans(),
            stable=st.booleans(),
        ),
        label="kw",
    )
```

~~Probably enforces there are always kwargs, so we don't test default values to behave as expected...~~

Edit: this isn't true, missing kwargs are tested by `hypothesis`.

The problem comes from the fact that **tested arrays are too small.**

It's a common optimization in sorting algorithms to switch to insertion sort for small arrays, and insertion sort is inherently stable. See `#define SMALL_QUICKSORT 15` constant in [numpy's code](https://github.com/numpy/numpy/blob/main/numpy/_core/src/npysort/quicksort.cpp) for instance.

More generally, it's common to switch strategy for small arrays in sorting algorithm, so it's good to test long enough arrays when testing sort/argsort.

cakedev0:
I would have been happy to write the PR, but the heavy use of `hypothesis` makes it hard for me, I don't know this lib.

PR #390, opened 2025-10-19
FIX: larger inputs for sorting tests - to test sorts stability

cakedev0:
Fixes: https://github.com/data-apis/array-api-tests/issues/389

To test sort stability of array libraries, you usually need arrays of more than 16 elements (because for less than 16, libs often default to an inherently stable sort, even for unstable algorithms).

Because of this was not tested properly, a gap in array-api-compat was not detected by this test suite. See issue https://github.com/data-apis/array-api-compat/issues/354. You can try running:
```bash
ARRAY_API_TESTS_VERSION="2024.12" ARRAY_API_TESTS_MODULE=array_api_compat.torch pytest array_api_tests/test_sorting_functions.py
```
with this branch of `array-api-tests` to see it indeed fails.

### DIFF
--- a/array_api_tests/test_sorting_functions.py
+++ b/array_api_tests/test_sorting_functions.py
@@ -34,6 +34,8 @@ def assert_scalar_in_set(
     x=hh.arrays(
         dtype=hh.real_dtypes,
         shape=hh.shapes(min_dims=1, min_side=1, max_side=50),
+        # argsort needs to be tested on big enough arrays, hence we set `max_side=50`
+        # see https://github.com/data-apis/array-api-tests/issues/389
         elements={"allow_nan": False},
     ),
     data=st.data(),
@@ -95,6 +97,8 @@ def test_argsort(x, data):
     x=hh.arrays(
         dtype=hh.real_dtypes,
         shape=hh.shapes(min_dims=1, min_side=1, max_side=50),
+        # sort needs to be tested on big enough arrays, hence we set `max_side=50`
+        # see https://github.com/data-apis/array-api-tests/issues/389
         elements={"allow_nan": False},
     ),
     data=st.data(),


ev-br:
Thanks for the explanation!
Indeed, TIL that the default argsorting strategy in pytorch differs between 16 and 17 elements:

```
In [1]: import torch

In [2]: t = torch.zeros(16)

In [3]: torch.argsort(t)
Out[3]: tensor([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15])

In [4]: t = torch.zeros(17)

In [5]: torch.argsort(t)
Out[5]: tensor([ 8, 16, 15, 14, 13, 12, 11, 10,  9,  0,  7,  6,  5,  4,  3,  2,  1])
```

> Note: The test would be much more explicit without using hypothesis, it would look like:
assert xp.all(xp.argsort(xp.asarray([0] * 30)) == xp.arange(30))
Maybe I can write a simple regression test like this?

That would be nice indeed. We tend to keep library-specific tests for behaviors which differ between  "bare" and array-api-compat wrapped libraries in array-api-compat itself, so this test would be a great addition to 
https://github.com/data-apis/array-api-compat/blob/main/tests/test_torch.py
In https://github.com/data-apis/array-api-compat/pull/356, which needs a small tweak anyway.

cakedev0:
> the default argsorting strategy in pytorch differs between 16 and 17 elements:

Note that it's probably a common trait of sorting algorithms, in numpy:
```Python
np.argsort(np.zeros(16, dtype='int8'))  # stable
np.argsort(np.zeros(17, dtype='int8'))  # unstable
```
(you don't see this behavior for dtypes that benefit from SIMD optimization though)

It seems that it's a common optimization to switch to insertion sort for small arrays, and insertion sort is inherently stable. See https://github.com/numpy/numpy/blob/main/numpy/_core/src/npysort/quicksort.cpp#L70 in numpy.

cakedev0:
I agree this would be quite hard to understand for future readers. I added a short comment that reference the issue #389, that should already be much easier to understand now.

PR #398, opened 2025-11-24
Fix testing on python 3.10

ev-br:
Implement the suggestion from https://github.com/HypothesisWorks/hypothesis/issues/4606#issuecomment-3568140174

Note: this PR is fully drafted by Copilot. The full history of prompts is at https://github.com/ev-br/array-api-tests/pull/2
My act is only to provide prompts and review the resulting PR.

The patch itself is fairly trivial. The trick is that it's a first patch prepared by a bot. So this should probably serve as a blueprint of how we want to deal with bot-authored commits.
Yay/nay @rgommers @kgryte @asmeurer ?

If we go with it, my plan is to squash-merge the PR, edit the commit message and keep copilot's authorship.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,8 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.12", "3.13"]
-
+        python-version: ["3.10", "3.12", "3.13", "3.14"]
     steps:
     - name: Checkout array-api-tests
       uses: actions/checkout@v1


ev-br:
To clarify my position: I think it's OK to merge this kind of bot-generated patches, provided
- they are reviewed with the same level of scrutiny as PRs from a human I never met behind a github username I never seen.
- The reviewer information is somehow recorded in the commit message. Specific way TBD: reviewed-by or signed-off-by from one of us with the commit bit.

The remaining concern is around copyright, as detailed at length at the [Discourse thread](https://discuss.scientific-python.org/t/a-policy-on-generative-ai-assisted-contributions/1702/28).

PR #405, opened 2026-01-10
ENH: expand testing of `meshgrid`

ev-br:
- draw the `indexing` argument
- check shapes of the outputs

This should have caught simple errors like https://github.com/data-apis/array-api-compat/pull/341


ev-br:
Tested locally with 10_000 examples for numpy, torch, jax.numpy and cupy; with 1000 examples for dask.array.

PR #409, opened 2026-01-16
ENH: parse complex special cases from stubs of unary functions

ev-br:
"stub" docstrings include "special cases":

> `if x_i is +infinity, sqrt(x_i) is +infinity` etc

`test_special_cases` parses these statements from docstrings and makes them into tests cases.

So far, parsing only worked for real value cases, and failed for complex-valued cases:

> For complex floating-point operands, let a = real(x_i), b = imag(x_i), and
> `If a is either +0 or -0 and b is +0, the result is +0 + 0j.`

These stanzas simply generate "case for {func} is not machine-readable" `UserWarning`s. Quite a wall of them.

Therefore, we update parsing and testing code to take these complex-valued cases into accout.
For now, we only consider unary functions.

The effect is that

`$ ARRAY_API_TESTS_MODULE=array_api_compat.torch pytest array_api_tests/test_special_cases.py::test_unary`

generates
- "128 passed, 177 warnings in 0.78s" on master
- "11 failed, 241 passed, 49 warnings in 1.82s" on this branch

So that there are new failures (from new complex-valued cases) but we have 128 less warnings.

-------------------

NB: this pull request is generated by Copilot. The original prompt and subsequent back-and-forths are at https://github.com/ev-br/array-api-tests/pull/3

### DIFF
--- a/array-api-strict-skips.txt
+++ b/array-api-strict-skips.txt
@@ -46,4 +46,7 @@ array_api_tests/test_special_cases.py::test_unary[expm1((real(x_i) is +0 or real
 array_api_tests/test_special_cases.py::test_unary[sign((real(x_i) is -0 or real(x_i) == +0) and (imag(x_i) is -0 or imag(x_i) == +0)) -> 0 + 0j]
 array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +infinity and isfinite(imag(x_i)) and imag(x_i) > 0) -> 1 + 0j]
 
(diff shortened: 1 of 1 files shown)


rgommers:
> NB: this pull request is generated by Copilot. The original prompt and subsequent back-and-forths are at [ev-br#3](https://github.com/ev-br/array-api-tests/pull/3)

Thanks @ev-br. Since it's the first such PR, I had a look - in my opinion this declaration and the level of visibility on how this code was generated is nice, and acceptable for this project. Given that this project is "advanced plumbing", I don't have any concerns about copyright, or extra review load from AI-generated contributions (you did the steering and review already).

PR #412, opened 2026-01-30
BUG: make test_repeat spec-compliant

ev-br:
Its value testing was using xp.full with array fill value, while only python scalars are allowed. Thus, convert fill_value to scalar.

cross-ref https://github.com/data-apis/array-api-strict/pull/181


lucascolley (approved):
thanks Evgeni!

Issue #194, opened 2023-08-08
test_nonzero in test_searching_functions.py incorrectly handle 0d inputs


oleksandr-pavlyk:
`xp.nonzero` is [documented in the spec](https://data-apis.org/array-api/latest/API_specification/generated/array_api.nonzero.html) to require an exception for 0d arrays:

> x (array) – input array. Must have a positive rank. If x is zero-dimensional, the function must raise an exception.

Yet the code exercises `xp.nonzero` for zero-dimensional arrays and expects the function to returns a tuple of size 1: https://github.com/data-apis/array-api-tests/blob/f82c7bc8627cc2c3a44fa3e425f53a253a609aa8/array_api_tests/test_searching_functions.py#L93-L98

This discrepancy should be fixed, or the spec modified.

PR #419, opened 2026-02-08
MAINT: remove a dead code branch from test_nonzero

ev-br:
The branch is checking for x.ndim == 0, while the hypothesis decorator only generates arrays of at least ndim=1:

`given(hh.arrays(dtype=hh.all_dtypes, shape=hh.shapes(min_dims=1, min_side=1)))`

This is consistent with the spec:
https://data-apis.org/array-api/draft/API_specification/generated/array_api.nonzero.html

> x (array) – input array. **Must** have one or more dimensions.
> If x is zero-dimensional, the function **must** raise an exception.

closes https://github.com/data-apis/array-api-tests/issues/194


ev-br:
tested locally with cupy/torch/numpy

Issue #108, opened 2022-03-23
Test `copy` arg in `reshape()`


honno:

asmeurer:
I'm implementing this in the numpy_array_api_compat library. A real test would be nice. It shouldn't be difficult to do. If the new shape is different, just check that the returned array has a different shape. If it's the same, we can do a check similar to the one in test_asarray. The None case cannot be tested because it depends on implementation details (like if the underlying library uses strides/views).

PR #432, opened 2026-03-22
ENH: reshape: test copy= kwarg

ev-br:
Apparently, `copy=` kwarg of `reshape` is <s>not very well</s> reasonably well supported:

- [x] torch does not have it (https://docs.pytorch.org/docs/stable/generated/torch.reshape.html), and array-api-compat simply raises -- fixed by https://github.com/data-apis/array-api-compat/pull/457
- [x] cupy will have it soon-ish (in cupy==14.0 via https://github.com/cupy/cupy/pull/10147); for now, there's a workaround in `array_api_compat`
- [x] jax.numpy has it but always copies anyway (unless jit elides the copy)

Where the argument is supported, the test itself needs work:
- [x] on `-strict`, zero-sized arrays fail some internal check with `copy=True` -- fixed by https://github.com/data-apis/array-api-strict/pull/232
- [x] on `numpy`, the test generates arrays or numpy scalars, and scalars choke something in the test itself. -- fixed by https://github.com/data-apis/array-api-compat/pull/454

### DIFF
--- a/array_api_tests/test_manipulation_functions.py
+++ b/array_api_tests/test_manipulation_functions.py
@@ -428,13 +428,15 @@ reshape_shape = st.shared(hh.shapes(), key="reshape_shape")
 @given(
     x=hh.arrays(dtype=hh.all_dtypes, shape=reshape_shape),
     shape=hh.reshape_shapes(reshape_shape),
+    kw=hh.kwargs(copy=st.booleans())
 )
-def test_reshape(x, shape):
-    repro_snippet = ph.format_snippet(f"xp.reshape({x!r},{shape!r})")
+def test_reshape(x, shape, kw):
+    repro_snippet = ph.format_snippet(f"xp.reshape({x!r},{shape!r}, **kw) with {kw =}")
     try: 
-        out = xp.reshape(x, shape)
+        out = xp.reshape(x, shape, **kw)
 
         ph.assert_dtype("reshape", in_dtype=x.dtype, out_dtype=out.dtype)
+        # TODO: test copy
 
         _shape = list(shape)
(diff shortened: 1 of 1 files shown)


ev-br:
cross-ref https://github.com/data-apis/array-api-tests/issues/108

ev-br:
Cupy issue to add the copy= argument to reshape: https://github.com/cupy/cupy/issues/10039

ev-br:
The array-api-strict failure seems to be 

```
(Pdb) l
118  	
119  	    data = x._array
120  	    if copy:
121  	        data = np.copy(data)
122  	
123  ->	    reshaped = np.reshape(data, shape)
124  	
125  	    if copy is False and not np.shares_memory(data, reshaped):
126  	        raise AttributeError("Incompatible shape for in-place modification.")
127  	
128  	    return Array._new(reshaped, device=x.device)
(Pdb) p data
array([], dtype=float64)
(Pdb) p np.shares_memory(data, data)
False
```

which a kind of makes sense: if there's no data to share, nothing's shared. So it'll need a tweak in `array-api-strict`.

ev-br:
It looks like that after https://github.com/data-apis/array-api-strict/pull/232 and https://github.com/data-apis/array-api-compat/pull/454 by  @stanbot8, the only remaining blocker is torch.

ev-br:
All wrapped modules have been fixed (thanks @stanbot8!), the updated test passes locally. We'll be able to improve further when
- numpy 2.1 is the minimum supported version (will be able to drop branching on the numpy version and always use the native numpy function)
- cupy 14 is the minimum supported version (will be able to drop the wrapper and use cupy's native `copy=`).

PR #433, opened 2026-03-23
ENH: Add basic `dlpack` testing

ev-br:
Add a basic set of `dlpack` tests. The tests here are "boring": 
- no cross-library tests, only test `from_dlpack` for a single array library
- only test `copy=False` on the same device
- <s>no tests for `__dlpack__` yet</s>  several `XXX` comments to mark where I've no idea.

### DIFF
--- a/array_api_tests/test_dlpack.py
+++ b/array_api_tests/test_dlpack.py
@@ -1,0 +1,127 @@
+from enum import Enum
+
+from hypothesis import given, strategies as st
+from . import _array_module as xp
+from . import pytest_helpers as ph
+from . import hypothesis_helpers as hh
+
+# dlpack Enum values,
+# https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack_device__.html
+
+class DLPackDeviceEnum(Enum):
+    CPU = 1
+    CUDA = 2
+    CPU_PINNED = 3
+    OPENCL = 4
+    VULKAN = 7
+    METAL = 8
+    VPI = 9
+    ROCM = 10
+    CUDA_MANAGED = 13
+    ONE_API = 14
+
+
+def _compatible_devices(devices):
+    """Given a list of devices, filter out dlpack-incompatible ones."""
+    # XXX: there seems to be no better way than try-catch for __dlpack_device__()
+
+    # XXX: this process actually fails with CuPy because CuPy ignores the device= argument
+    # cf https://github.com/data-apis/array-api-compat/issues/337 and
+    # https://github.com/cupy/cupy/issues/9848
+    # Luckily, CuPy only supports CUDA devices, and they are all compatible.
+    compatible_ = []
+    for device in devices:
+        x = xp.empty(2, device=device)
+        try:
+            x.__dlpack_device__()
+        except:
+            # case in point: torch.device(type="meta") raises
+            # ValueError: Unknown device type meta for Dlpack
+            pass
+        else:
+            # no exception => device is compatible
+            compatible_.append(device)
+    return compatible_
+
+
+@given(dtype=hh.all_dtypes, data=st.data())
+def test_dlpack_device(dtype, data):
+    """Test the array object __dlpack_device__ method."""
+    # TODO: 1. generate inputs on non-default devices
+    x = xp.empty(3, dtype=dtype)
+    device_type, device_id  = x.__dlpack_device__()
+
+    assert DLPackDeviceEnum(int(device_type))
+    assert isinstance(device_id, int)
+
+
+@given(
+    x=hh.arrays(dtype=hh.all_dtypes, shape=hh.shapes(min_dims=1, max_side=2)),
+    copy_kw=hh.kwargs(
+        copy=st.booleans() | st.none()
+    ),
+    max_version_kw=hh.kwargs(
+        max_version=st.tuples(
+            st.integers(min_value=0, max_value=2),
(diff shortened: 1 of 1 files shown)


ev-br:
Issues with these "trivial" tests:

1. `torch` has a weird `device(type="meta")` which is incompatible with DLpack. If anything, the following should raise a `BufferError` IIUC,  not a `ValueError`

```
In [8]: device = torch.device(type="meta")

In [9]: x = torch.empty(3, device=device)

In [10]: torch.from_dlpack(x)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 torch.from_dlpack(x)
...
ValueError: Unknown device type meta for Dlpack
```

2. `jax` fails with `copy=True` on CUDA where it works on CPU:

```
In [8]: jnp.from_dlpack(jnp.ones(3), copy=True)
---------------------------------------------------------------------------
JaxRuntimeError                           Traceback (most recent call last)
Cell In[8], line 1
----> 1 jnp.from_dlpack(jnp.ones(3), copy=True)

...

JaxRuntimeError: UNIMPLEMENTED: PJRT C API does not support HostBufferSemantics other than HostBufferSemantics::kImmutableOnlyDuringCall, HostBufferSemantics::kImmutableZeroCopy and HostBufferSemantics::kImmutableUntilTransferCompletes

```

3. `cupy==14.0.1` doesn't implement `copy` or `device` arguments:

```
>>> x =  cupy.array([], dtype=float)
>>> cupy.from_dlpack(x, copy=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/_core/dlpack.pyx", line 568, in cupy._core.dlpack.from_dlpack
  File "cupy/_core/core.pyx", line 355, in cupy._core.core._ndarray_base.__dlpack__
BufferError: copy=True only supported for copy to CPU.

>>> cupy.from_dlpack(x, device=x.device)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/_core/dlpack.pyx", line 554, in cupy._core.dlpack.from_dlpack
NotImplementedError: from_dlpack() does not support device yet.

>>> cupy.__version__
'14.0.1'
```

4. `numpy` cannot consume a capsule created by `__dlpack__`, while torch happily does. Which one is correct?

```
>>> import numpy as np
>>> import torch
>>> x = np.ones(3)
>>> capsule = x.__dlpack__()
>>> torch.from_dlpack(capsule)             # works fine
tensor([1., 1., 1.], dtype=torch.float64)
>>> np.from_dlpack(capsule)                 # ouch
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'PyCapsule' object has no attribute '__dlpack__'
```

ev-br:
Larger issues:

5. How to tell if a device is dlpack-compatible. The only way I see now is to try-except call `from_dlpack` and catch a BufferError.

6. How to tell if a `copy=False` device transfer is meant to work, given `source_device` and `target_device`. A first reaction is to create a dummy array on each device and check their `__dlpack_device__` values. If they match, it should work, ok. But then what about `CUDA`  vs `CUDA_MANAGED`? `CPU` vs `CPU_PINNED`? Other values? It feels like there needs to be something in the inspection capabilities to tell it.

7. How to tell a supported `max_version` in `__dlpack__`. Should `max_version=(-111, 42)` raise? What about reasonable version values: `max_version=(3, 0)` if a library only supports `(2,0)`.

rgommers:
(2) JAX doesn't implement DLPack 1.0 yet

(4) that shouldn't be tested, passing in a PyCapsule is not regular usage. 

(5) probably a try-except indeed, given that not every device of every library will be compatible. seems fine to do it like that. I'd expect `__dlpack_device__` to also raise then, but not sure all libraries actually do that.

(6) I think only matching devices are guaranteed to work, anything else probably won't but there are likely specific combos that will. Which seems fine (there's nothing in the standard that says it's not fine)?

(7) Whether an unreasonable value raises and what exception type it raises is almost always undefined in the standard. No need to test I'd say. For reasonable values: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html seems unambiguous. If the consumer say `(3, 0)` and the producer can only do `(2, 0)`, then it should do that and the consumer then may or may not be able to handle that. If it can't, it'll raise a `BufferError`.

pearu:
> (2) JAX doesn't implement DLPack 1.0 yet

While this is true (JAX officially supports 0.8, although it could support also 1.0 as the latest XLA ships with dlpack 1.1), the underlying reason for failure in (2) is likely different. For instance, (2) worked fine in past:
```python
>>> import jax.numpy as jnp
>>> x=jnp.from_dlpack(jnp.ones(3), copy=True)
>>> x.device
CudaDevice(id=0)
>>> jax.__version__
'0.8.1.dev20250703+2ac4523a7'
```

ev-br:
Thanks @pearu --- so (2) is essentially https://github.com/jax-ml/jax/issues/33790. 
I tested with `jax.__version__ == 0.9.2`,  the jax issue lists `0.8.1`, and your snippet has an earlier dev version in the 0.8.1 time frame IIUC. Meaning, it got broken around 0.8.1 and not fixed yet.

ev-br:
Landed this to get _some_ testing going. We'll be able to track improvements through the umbrella issue, https://github.com/data-apis/array-api-tests/issues/301

PR #445, opened 2026-07-28
CI: use array_api_strict main branch for testing

ev-br:
On CI, use the `array_api_strict` main branch instead of a released version. That way, when something is fixed in -strict, 
we don't have to wait for a release to unblock the CI here.


ev-br:
Let's give it a shot and test-drive in https://github.com/data-apis/array-api-tests/pull/432 which needs https://github.com/data-apis/array-api-strict/pull/232 to pass.

PR #448, opened 2026-08-05
fix: avoid dtype comparison for None report metadata

### LINKED ISSUE #447: None dtype comparison not supported in MLX
In reporting.py we have
```python
def to_json_serializable(o):
    if o in dtype_to_name:
        return dtype_to_name[o]
```
While running tests all the test failed because this method was being called at each test and the main problem is if the `o` is of the type `None`

MlLX throws an exception for equality like 
```python
if None == mlx.core.float16:
    print("success")
```

We either have to modify the method to something like

```python
def to_json_serializable(o):
    if o is not None and o in dtype_to_name:
        return dtype_to_name[o]
```

I think it's a minor problem and can be fixed in test itself not sure though.

tomatotomata:
## Summary

- avoid comparing `None` with array-library dtype objects before looking them up in `dtype_to_name`
- add a regression test using an MLX-like dtype whose equality operation rejects `None`

Closes #447

## Validation

- `ARRAY_API_TESTS_MODULE=numpy PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q -p pytest_jsonreport.plugin meta_tests/test_reporting.py` passed: 1 test
- `ARRAY_API_TESTS_MODULE=numpy PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q -p pytest_jsonreport.plugin meta_tests/test_utils.py` passed: 30 tests
- `python -m compileall -q reporting.py meta_tests/test_reporting.py` passed
- `git diff --check` passed

The full suite was not run. The focused commands disable unrelated globally installed pytest plugins and load the repository's JSON-report plugin explicitly.

### DIFF
--- a/meta_tests/test_reporting.py
+++ b/meta_tests/test_reporting.py
@@ -1,0 +1,16 @@
+from array_api_tests.dtype_helpers import EqualityMapping
+import reporting
+
+
+class MLXDtype:
+    def __eq__(self, other):
+        if other is None:
+            raise TypeError("MLX dtype equality does not support None")
+        return self is other
+
+
+def test_to_json_serializable_none_does_not_compare_to_dtypes(monkeypatch):
+    dtype = MLXDtype()
+    monkeypatch.setattr(reporting, "dtype_to_name", EqualityMapping([(dtype, "float16")]))
+
+    assert reporting.to_json_serializable(None) is None
--- a/reporting.py
+++ b/reporting.py
@@ -17,7 +17,7 @@ except ImportError:
     raise ImportError("pytest-json-report is required to run the array API tests")
 
 def to_json_serializable(o):
-    if o in dtype_to_name:
+    if o is not None and o in dtype_to_name:
         return dtype_to_name[o]
     if isinstance(o, (BuiltinFunctionType, FunctionType, type)):
         return o.__name__


ev-br:
I've to admit this looks a bit overboard. If the problem is that `None in dtype_to_name` fails with `mlx.core`, would the following do the trick? cc @aaishwarymishra --- would it fix the problem you were having:

```diff
$ git diff
diff --git a/reporting.py b/reporting.py
index 579aa21..0dcf410 100644
--- a/reporting.py
+++ b/reporting.py
@@ -17,6 +17,8 @@ except ImportError:
     raise ImportError("pytest-json-report is required to run the array API tests")
 
 def to_json_serializable(o):
+    if o is None:
+        return o
     if o in dtype_to_name:
         return dtype_to_name[o]
     if isinstance(o, (BuiltinFunctionType, FunctionType, type)):

```

ev-br:
Actually, scratch that. Need some more coffee, sorry. Now that I re-read the function itself, my diff above is the same fix. 
LGTM, thanks @tomatotomata

aaishwarymishra:
@ev-br happens to the best of us 😭

PR #451, opened 2026-08-12
BUG: use Python integers for repeat slice bounds

### LINKED ISSUE #450: Bug: test_repeat uses the 0 dim array for slicing instead of ints
`test_repeat ` in the `array_api_tests/test_manipulation_functions.py`,  the test creates an MLX array of repeat counts:

```python
repeats_array = xp.full(size, repeats, dtype=xp.int32)
```

Iteration returns a 0D MLX array, so `end` is also an MLX array:

```python
end = start + count
out_slice[start:end]  # ValueError: Slice indices must be integers or None
```

Convert `count` to a Python integer before calculating the slice:

```python
count = int(count)
end = start + count
```

complete loop code 
```python
for i, count in enumerate(repeats_array):
    end = start + count
    x_slice_i = dh.get_scalar_type(x.dtype)((x_slice[i]))
    ph.assert_array_elements("repeat", out=out_slice[start:end],
                             expected=xp.full((int(count),), x_slice_i, dtype=x.dtype),
                             kw=kw)
    start = end
```

Boulea7:
## Summary

Closes #450. `test_repeat` now converts each array-valued repeat count to a Python integer before computing slice bounds. This keeps the value check compatible with libraries such as MLX that do not accept 0D arrays as slice indices.

## Validation

- Reproduced the failure with an integer-convertible scalar that does not implement `__index__`, then confirmed the same path passes after the change
- `ARRAY_API_TESTS_MODULE=array_api_strict .venv/bin/python -m pytest -q array_api_tests/test_manipulation_functions.py::test_repeat --max-examples=1000 --disable-deadline`
- `ARRAY_API_TESTS_MODULE=array_api_strict .venv/bin/python -m pytest -q meta_tests` (70 passed, 1 skipped)
- `.venv/bin/pre-commit run --all-files`

The full manipulation test file also surfaced an unrelated `test_reshape` failure for an empty array with `copy=False`; the changed `test_repeat` passes independently.

### DIFF
--- a/array_api_tests/test_manipulation_functions.py
+++ b/array_api_tests/test_manipulation_functions.py
@@ -410,10 +410,11 @@ def test_repeat(x, kw, data):
             out_slice = out[idx]
             start = 0
             for i, count in enumerate(repeats_array):
+                count = int(count)
                 end = start + count
                 x_slice_i = dh.get_scalar_type(x.dtype)((x_slice[i]))
                 ph.assert_array_elements("repeat", out=out_slice[start:end],
-                                         expected=xp.full((int(count),), x_slice_i, dtype=x.dtype),
+                                         expected=xp.full((count,), x_slice_i, dtype=x.dtype),
                                          kw=kw)
                 start = end
 


ev-br (approved):
LGTM, thanks

PR #454, opened 2026-08-13
Add supportfor multiple skip/xfail files

betatim:
This allows you to specify several files which are combined. For examplea general file plus a platform specific one.

You can now do something like `pytest --skips-file skips-general.txt --skips-file skips-macos.txt array_api_tests/`.

There was a difference in line handling between skips and xfails. I unified them to how xfails handles things. The skips lines were only taken `if line.startswith("array_api_tests")`. And all other lines were ignored. Now all lines that aren't blank or start with a `#` are kept. They still have to resolve to valid test IDs, so I think the end result will be the same.

This came out of a discussion at https://github.com/data-apis/array-api-strict/pull/240#discussion_r3765157360

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,14 @@ By default, the skips and xfails files are `skips.txt` and `fails.txt` in the ro
 of this repository, but any file can be specified with the `--skips-file` and
 `--xfails-file` command line flags.
 
+Both flags can be given several times, in which case the files are merged. This
+is useful to keep entries which only apply to some platforms separate from the
+general ones:
+
+```
+pytest --skips-file skips-general.txt --skips-file skips-macos.txt array_api_tests/
+```
+
 The files should list the test ids to be skipped/xfailed. Empty lines and
 lines starting with `#` are ignored. The test id can be any substring of the
 test ids to skip/xfail.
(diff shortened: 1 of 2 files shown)


betatim:
I'm not sure who should/can review this, so I requested a review from @ev-br because GitHub suggested it.

How do people generally decide on merging (or not) things in this repo? Wait for one review? One thumbs up from Lucas and then self-merge?

ev-br:
Test-driving this PR locally: 

1. The following incantation fails on master and passes in this PR. This is the expected correct behavior: the test is xfailed in the `numpy-xfails.txt`, so in master two xfail files are not understood, and in this PR they are.

```
$ ARRAY_API_TESTS_MODULE=numpy pytest array_api_tests/test_data_type_functions.py::test_finfo[float32]  --skips-file=../array-api-compat/numpy-xfails.txt --skips-file=../array-api-strict/array-api-tests-xfails.txt  -v
```

2. Running with no skips skill works: the following incantation fails as expected

```
$ ARRAY_API_TESTS_MODULE=numpy pytest array_api_tests/test_data_type_functions.py::test_finfo[float32] 
``` 

The code change LGTM, thank you @betatim

ev-br:
> How do people generally decide on merging (or not) things in this repo? Wait for one review? One thumbs up from Lucas and then self-merge?

For the last year or so, it was mostly me self-merging at ad hoc times. 
Not that it's healthy or desired, and I'd be very happy for more hands and eyes :-).

PR #462, opened 2026-09-05
Fix matrix transpose reference construction

### LINKED ISSUE #461: `test_linalg.test_matrix_transpose()` calls `asarray` on nested sequence of arrays
According to the standard, `asarray` must support nested sequences of Python scalars. However, `test_matrix_transpose` ultimately calls `asarray` on a list of lists of arrays:

https://github.com/data-apis/array-api-tests/blob/434a419ba3204ff9b67a8df9e2162998877d5935/array_api_tests/test_linalg.py#L516-L522

Crossref gh-142.

mdhaber:
`_test_matrix_transpose` constructs its expected result by passing nested lists of zero-dimensional arrays to `asarray`, although the standard only requires nested sequences of Python scalars. Construct the expected transpose by stacking the input rows along axis 1. For empty matrices, allocate the transposed shape explicitly with the input dtype and device, since `stack` requires a nonempty sequence. This also preserves both dimensions for empty matrices.

Add regression coverage that rejects non-Python-scalar sequence elements in `asarray`, covering rectangular matrices, stacks, and empty dimensions.

Validation with array-api-strict 2.6.1 on Python 3.14:
- Confirmed the new regression cases fail before the fix (nested-array inputs and lost empty dimensions).
- Both matrix transpose tests and all linalg meta-tests pass (8 passed).
- Flake8 `--select F` on both changed files and `git diff --check` pass.

Fixes #461.

### DIFF
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -516,10 +516,14 @@ def test_matrix_rank(x, kw):
 def _test_matrix_transpose(namespace, x):
     matrix_transpose = namespace.matrix_transpose
     res = matrix_transpose(x)
-    true_val = lambda a: _array_module.asarray([[a[i, j] for i in
-                                                range(a.shape[0])] for j in
-                                                range(a.shape[1])],
-                                               dtype=a.dtype)
+    def true_val(a):
+        expected = _array_module.empty((a.shape[1], a.shape[0]),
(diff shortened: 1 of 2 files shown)


mdhaber:
This is the ai slop version of gh-463, independently written and posted by GPT-6 Atra (medium effort) via Codex.

PR #463, opened 2026-09-05
test_matrix_transpose: avoid calling asarray on nested sequence of arrays

### LINKED ISSUE #461: `test_linalg.test_matrix_transpose()` calls `asarray` on nested sequence of arrays

mdhaber:
Closes gh-461

### DIFF
--- a/array_api_tests/test_linalg.py
+++ b/array_api_tests/test_linalg.py
@@ -516,9 +516,9 @@ def test_matrix_rank(x, kw):
 def _test_matrix_transpose(namespace, x):
     matrix_transpose = namespace.matrix_transpose
     res = matrix_transpose(x)
-    true_val = lambda a: _array_module.asarray([[a[i, j] for i in
-                                                range(a.shape[0])] for j in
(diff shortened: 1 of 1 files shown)


ev-br (approved):
Ran `$ ARRAY_API_TESTS_MODULE=numpy pytest array_api_tests/test_linalg.py::test_matrix_transpose -v  --max-examples=50_000` to triple-check that using `stack` does not throw off numpy's indexing returning scalars not 0D arrays, and it passed. Thus all is well, and the fix is clearly improves the Array API compatibility. 

Merging, thanks @mdhaber. 

I'll send a follow-up to clean up the duplication of
```
from . import _array_module
from . import _array_module as xp
```

PR #469, opened 2026-09-15
TST: linalg: avoid numerical difficulties in tests involving matrix inversion

### LINKED ISSUE #467: `hypothesis_helpers.invertible_matrices` generates numerically singular matrices
`invertible_matrices` is:

https://github.com/data-apis/array-api-tests/blob/ef5b39f1190d54dca0d9a12646033d2d123afdd5/array_api_tests/hypothesis_helpers.py#L426-L443

There is no upper bound on the magnitude of the entries, so it is free to generate (and reliably generates, in my experience) very poorly conditioned matrices that `mpmath` considers to be numerically singular.

The comment says that it is trying to avoid numerical issues; to do that, it would need to bound the relative magnitudes of the entries. 

The assertion `xp.all(xp.abs(d) >= 0.5)` can also be removed since that is guaranteed by the parameters provided to `one_of`.

mdhaber:
Closes gh-467.


array_api_tests/hypothesis_helpers.py
```diff
@@ -428,17 +428,14 @@ def invertible_matrices(draw, dtypes=floating_dtypes, stack_shapes=shapes()):
...
+                   allow_nan=False, allow_infinity=False),
+        from_dtype(dtype, max_value=-0.5, min_value=-50,
+                   allow_nan=False, allow_infinity=False),
     )
     d = draw(arrays(dtype, shape=(*stack_shape, 1, n), elements=elements))
 
-    # Functions that require invertible matrices may do anything when it is
-    # singular, including raising an exception, so we make sure the diagonals
-    # are sufficiently nonzero to avoid any numerical issues.
-    assert xp.all(xp.abs(d) >= 0.5)
```
mdhaber, line 440: `xp.all(xp.abs(d) >= 0.5)` was guaranteed by the definition of `elements`, yet this alone did not avoid numerical issues, so I went ahead and removed it. With the change above, I think the worst condition number possible is 100.

ev-br:
Thanks

